PR #13, opened 2022-02-23
Update notebook and api for fitter

weiliangjin2021:


tylerflex:
Hey @weiliangjin2021  can we add one thing?  After the wvl_range is set, how do I "unset" it back to the original data?  Could we just add a mention? Otherwise looks great and I'll merge it in when that's done.

weiliangjin2021:
I added a comment in that part that it can unset by setting the values to None.

tylerflex:
Thanks :)

PR #14, opened 2022-02-23
added scattering matrix notebook

tylerflex:


shashwat-sh (approved):
Looks good! The arrows don't seem to show up when plotting the ports - is that intentional, or just that those changes aren't included in this branch? If the latter, might be worth re-running this notebook after merging with develop using the latest tidy3d updates.

tylerflex:
I think it's because of the submodule branch yea, I'll figure it out when I merge and run the notebooks again. thank you for the reviews!

tylerflex:
I grabbed the changed files from develop rather than merge.

PR #35, opened 2022-06-02
Add downsampling example for sphere RCS

shashwat-sh:
Introduced an additional downsampled near field monitor around the sphere. In the end, the RCS is compared to Mie results for both the original and downsampled monitors.

Feel free to ignore this PR and just manually copy over the one notebook if that's easier: `Near2FarSphereRCS.ipynb` is the only file that needs to be merged.


tylerflex:
thanks, merged

PR #46, opened 2022-09-23
Added diffraction notebook

shashwat-sh:


momchil-flex:
- There's a `DiffractionEfficiencyMonitor` mention in the text that should just be `DiffractionMonitor`
- Let's use `plot` instead of `plot_eps`, I think we'd rather direct our user towards that (e.g. with dispersive materials they tend to get confused)

momchil-flex:
```
monitor_r = td.DiffractionMonitor(
    center=[0., 0., length_z / 2 - wavelength * 0.5],
    size=[td.inf, td.inf, 0],
    freqs=[f0],
    name='reflection',
    orders_x = orders,
    orders_y = orders,
    normal_dir='-',
)

monitor_t = td.DiffractionMonitor(
    center=[0., 0., -length_z / 2 + wavelength * 0.5],
    size=[td.inf, td.inf, 0],
    freqs=[f0],
    name='transmission',
    orders_x = orders,
    orders_y = orders,
    normal_dir='-',
)
```

These should have opposite `normal_dir`, no? In our setup I would think the reflection one should be `+`?

PR #48, opened 2022-10-25
Tom/mmi splitter

tomflexcompute:
Added the MMI example model.


tylerflex (changes requested):
Hi Tom, this notebook looks great! Thanks for contributing it! Here are a few comments, feel free to reply here if you have questions or want to discuss more. If you are ready you can submit another commit to fix some of them and then let me know that you'd like another review. 

1. Please insert a line in the intro for running on binder, see other notebooks for example.
```
Run this notebook in your browser using [Binder](https://mybinder.org/v2/gh/flexcompute-readthedocs/tidy3d-docs/readthedocs?labpath=docs%2Fsource%2Fnotebooks%2FMMI_1by4.ipynb).
```

2. We may want to define some more geometric parameters just to be sure if we change anything it can still work properly. For example: define some spacing between pml and structure in x,y,z. Use for Lx, Ly, Lz calculation and monitor geometry. For example, we currently the `Ly =  L1+L_MMI+L2 + 3` where I assume the 3 is 2*spacing of PML. Then later on the center of the flux monitor is `Ly-1.5`. We should avoid hard coding the 3 and 1.5.  Is my understanding correct?
3. Let’s make `mode_index=0` a variable in the mode solver section and then pass it in as `mode_data.Ex.isel(mode_index=mode_index)` in all 3 of the calls, that way the user can quickly visualize different modes by changing the single variable.
4. After first MMI field plot, typo “outter" -> “outer” (same in the plot legend below).
5. Perhaps 3x3=9 parameter combinations is not quite enough? If we extend to, say 25 does it take significantly longer? I’d also be curious about doing, let’s say, a 1D scan of L_MMI (8 points) followed by a 1D scan of W_MMI (8 points) at the optimal L value and seeing how that performs? What are your thoughts? do you think 3x3 is sufficient or can we do better without too much slowdown by doing more points or by doing a few 1D scans at higher resolution?

tylerflex:
Ok, we can stick to a 2D scan. What do you think about 4x4? does it still look good and is it faster?

tomflexcompute:
4x4 would reduce the run time to about 10 mins. I think it looks fine too.

tomflexcompute:
The latest commit is with a 4x4 scan.

tylerflex:
Thanks, I think maybe for simplicity, we can revert to the 3x3 case. It seems the results were about the same and took shorter. What do you think?

tylerflex (approved):
Looks good! Feel free to merge into develop using the “rebase and merge” option.

PR #49, opened 2022-10-28
Adding gui example notebooks to main notebooks folder

shashwat-sh:
Adding these case studies to the notebooks folder and making sure they show up in the doc examples:
- Bragg grating
- Biosensor grating
- Plasmonic nanoparticle

Previously, these were stored in a separate "extra_examples" folder.


tylerflex:
Great! Should we make a new subsection for these you think? or do you think for now under case studies makes most sense?

tylerflex (commented):
## Plasmonic nanoparticles
* Add binder link
* "the rapid field variations inside and near the particles requires a fine local discretization" -> "require"
* are we confident that we can just subtract the (no particle) field from the total field to get the scattered field? should we comment on the validity of this approach?
* the grid plots look same left and right (small red thing in center) is this expected?
* Maybe some words on what explains difference between mie and tidy3d? do we expect it to go down with resolution?
* why no mie flux plotted?

## Bragg Grating
* looks good, not sure if any more discussion is required or useful at the very end. Something just discussing that from the plots we can see there's a band where one of the gratings successfully reflects the incident light?
* only thing I would say is that the "adding corregation" part is a bit complex so it might be better to have it in a function that gets called a few times instead of copied and pasted, but not a high priority.

## Plastic Biosensor Grating
* Also looks good, same thing where I dont know if it might make sense to add a short discussion at the end, saying for example that at the peak reflection frequency, the transmission of the grating is very sensitive to any changes and can therefore operate as a good biosensor.

tylerflex:
@caseyflex when you get time can you read through these notebooks and just note if you see any room for improvements? It will also be good for you to see a few more case studies for your tidy3d learning.

shashwat-sh:
> Great! Should we make a new subsection for these you think? or do you think for now under case studies makes most sense?

Maybe when the GUI project links are active we can move them to a "GUI projects" subsection?

shashwat-sh:
@tylerflex Thanks, I've made the changes you suggested. Yeah there was a definite need for explanations - I forgot that when I made these notebooks, it was when we planned to just use the resulting json for the GUI.

> why no mie flux plotted?

It's a bit of a pain to extract quantities at arbitrary observation points with the Mie library I was using, especially for near-field quantities. I can try to do this when I get a chance and add the results in.

tylerflex:
Yea sounds good

caseyflex (approved):
Looks good to me. Interesting set of examples!

tylerflex (approved):

PR #50, opened 2022-11-01
Tom/polarizer rotator

tomflexcompute:
This is the example model for an integrated photonic polarization splitter/rotator device.


tylerflex (approved):
Looks good to me! Just a few questions:
* from the simulation.plot, it seems the sources and monitors are extremely close to the boundaries. Is this expected? I dont even see the PML.  
image. Can you comment on why it looks strange? I guess it's just the aspect ratio?
* cell [10] contains an `ax = ax`, let's change to `ax=ax`.
* This is a pretty large simulation right? I wonder if we should highlight just how big it is (number of cells, approx how long would take to simulate in CPU, for example) after setting it up.

tomflexcompute:
Thanks @tylerflex for reviewing. 

Indeed the plot looks a little strange purely due to the large aspect ratio. The device is very long in the y direction so I don't plot it in this scale, we could barely see anything in the plot but a long thin rod. The thickness in all directions should be the same 

Good catch. This has been fixed.

This model is actually not that large compared to some of the larger models we have. It's long in y but very short in x and z so overall the volume is not very large. In Lumerical with a similar mesh, I think it probably takes 10 minutes as in Tidy3D it takes around 1 minute. A regular desktop should be able to run this model in Lumerical but then the speed depends on their CPU clock speed, number of cores, etc.

tylerflex:
ok thanks, feel free to merge this

PR #51, opened 2022-11-02
Updating autodoc pydantic configuration

momchil-flex:
These settings seem to clean up all the repeated information while preserving methods, etc.


tylerflex (approved):
with a comment, that I probably wouldn't remove but am ok if everyone else wants to.

momchil-flex:
Hm, this is true. I could be down to leave those in (while still keeping `autodoc_pydantic_model_show_field_summary = False`).

tylerflex:
fine by me

PR #52, opened 2022-11-02
Updating notebooks for updated near2far data interface

shashwat-sh:
This is just to keep the near2far-related notebooks up to date with changes to the data interface. Also added the diffraction monitor data structures to the API reference, which I had forgotten to do previously.


tylerflex:
looks good, only issues I see is:
* `Near2FarSphereRCS.ipynb` has a bunch of `Output()` lines, think we can remove them?
* `PlasmonicNanoparticle.ipynb` the Ez field plot has a lot of field inside of the nanoparticle. is this expected given that it's a conductor?

tylerflex (approved):
for fixing merge conflicts in api.rst, just keep the changes from your branch and overwrite the ones in develop. I deleted radiation vectors in develop because it wasn't linking.

shashwat-sh:
>    Near2FarSphereRCS.ipynb has a bunch of Output() lines, think we can remove them?

Done!

>    PlasmonicNanoparticle.ipynb the Ez field plot has a lot of field inside of the nanoparticle. is this expected given that it's a conductor?

The structure is very sub-wavelength in size, so the skin depth is pretty large at this frequency. I think the penetration of the E field inside is expected, e.g. https://optics.ansys.com/hc/en-us/articles/360042703433, the fields are expected to be strongest near the surface, but still don't necessarily decay quickly to 0 inside (note: those are E^2 plots).

PR #55, opened 2022-11-03
Tom/waveguide crossing

tomflexcompute:
This is an example of a waveguide crossing. An entry is added to the examples.rst and the schematic image is added to the img folder. @momchil-flex could you please review this?


momchil-flex:
Looks great. Minor comment: in the plot it seems like there's a tiny gap between the vertical tapers and the crossing region. I am 99% sure this is a plotting artifact but could you zoom in and make sure?

image

One other thing, if you don't want the whole structure to appear yellow, you could e.g. plot(z=0.1). Matter of personal preference though, I'm fine either way.

momchil-flex (approved):

tomflexcompute:
Thanks @momchil-flex for reviewing. You can be 100% sure that the gap is a plotting artifact. I tried to play with the size and resolution of the figure but there will always be some artifact in one or two of the four square edges... In the web GUI it looks perfectly fine. 

Actually I intentionally want to make sure the monitor is set up correctly at z=0 so this plot can not only verify the geometry but also make sure the monitor is in place.

momchil-flex:
Sounds good, go ahead and merge, thanks!

PR #65, opened 2022-11-12
Remove broken links to tidy3d-alpha doc

xin-flex:
from two places:
- Start Here page (quickstart.rst)
- Index page and sidebar (index.rst)


tylerflex:
@xin-flex @momchil-flex do we have any links to the alpha docs anymore?

PR #66, opened 2022-11-15
Tom/dielectric metamaterial

tomflexcompute:
Hi @momchil-flex . Could you please review this metasurface absorber notebook?


momchil-flex:
Looks good, thanks. Just two comments:

1. When you refer to Tidy3d components, it would be nicer to provide a link to the docs. So e.g. instead of `Medium.from_nk`, you could put:
```
The `from_nk` method of [Medium](https://docs.simulation.cloud/projects/tidy3d/en/latest/_autosummary/tidy3d.Medium.html)
```
You can similarly link the `Drude` medium. This comment also applies for future (and past...) notebooks. Generally, it would be good if we figure out a more automatic or stable way to link (e.g. this is a bit dangerous and the links may get broken because we're putting them in by hand and if something small changes in the url, it won't be automatically corrected). For now we haven't figured out a better way though.

2. Do you know how to squash your three commits into 1 before merging?

tomflexcompute:
Thanks @momchil-flex . Indeed it's a great idea to link the documentation whenever we mention something relevant. The notebook has been updated. 

I've also squashed all commits in this branch into one. Do you think everything is ok now?

momchil-flex:
Yep, merged - thanks!

PR #67, opened 2022-11-16
use adjoint plugin and add inverse design mode converter example

tylerflex:
Note: do not merge for 1.8


tylerflex:
closing as public adjoint notebooks are in #104

PR #72, opened 2022-11-24
Field projection notebook updates based on refactor and new features

shashwat-sh:
* Updated all field projection notebooks with new API and name changes

* Added exact projections for the metalens and zone plate

* Added a comprehensive field projection tutorial notebook

* Updated scattering cross section notebooks for clearer field subtraction


tylerflex (changes requested):
a few comments from the sphinx build:
* `Near2Far.ipynb` no longer exists, so let's remove it from `examples.rst`
* Looks like `FieldProjections.ipynb` and `ZonePlateFieldProjection.ipynb` are not added to `examples.rst`, should they be added?

### field projections
* the first image shows up a bit small in my notebook, not sure about in the build, but we may need to adjust a little.
* still set up to use your local solver, so will need to change to webapi.
* we might be able to leverage `xarray` [plotting](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.plot.pcolormesh.html) in the `make_field_plot` function rather than passing in `.values`. I it would would help users get more comfortable with `xarray` and make tidy3d look easier to use if we can include it.
* Consider formatting your f strings in the error percentage. I think we only need at most 2 decimal places (eg. `f"error = {error:.2f} %"`). This applies to other parts of the notebook as well.
* Consider adding a `plt.show()` after plotting stuff to avoid the extra text output in the notebooks.
* for timing, consider using the ipython magic command `%%time` instead of `time.time()`. As far as I remember, using `time.time()` isnt always the most reliable way to do it. For example, move this:
```py
import time
t0 = time.time()
projected_field_data = get_proj_fields(sim_data, monitor_near, monitor_far)
t1 = time.time()
proj_time = t1 - t0
```
to its own cell and use
```
%%time
projected_field_data = get_proj_fields(sim_data, monitor_near, monitor_far)
```
* Do you have a reference or formula for the analytical solution you're comparing against that you can include?
* More of an API concern, I'm not a big fan of `renormalize_fields` seemingly modifying the data contained in the object that calls it. It seems to break the immutability design we have going. I'd rather it returns a copy with the data renormalized @momchil-flex. Correct me if I'm mistaken.
* You know what's up with this warning?
```
/tmp/ipykernel_572536/2607939092.py:13: UserWarning: The input coordinates to pcolormesh are interpreted as cell centers, but are not monotonically increasing or decreasing. This may lead to incorrectly calculated cell edges, in which case, please supply explicit cell edges to pcolormesh.
  im = ax.pcolormesh(np.squeeze(phi), np.squeeze(theta) * 180 / np.pi, np.abs(Etheta), cmap='RdBu', shading='auto')
```
* I feel like maybe this notebook is so long, we should consider breaking it up into a separate Greens Function notebook and have the two notebooks reference each other. Any thoughts?

### in zone plate notebook
* the text:
```
For this example, we use `FieldProjectionCartesianData.fields_cartesian()` to get the fields at the previously-set `x,y,z` points.
``` 
is used.
Let's drop the `()` to avoid confusion because it seems like it's a property
* at the very end, we compute intensity using
```py
intensity = np.squeeze(
    np.square(np.abs(projected_fields['Ex'].values)) +\
    np.square(np.abs(projected_fields['Ey'].values)) +\
    np.square(np.abs(projected_fields['Ez'].values))
)
```
I feel like the fields should be co-located before computing this. Are they co-located? And if so, maybe what we can do is simply utilize the `xarray` functionality. For example:
```py
intensity = 0
for field in ('Ex', 'Ey', 'Ez'):
    intensity += abs(projected_fields[field])**2

_, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 5))
np.sqrt(intensity).plot(ax=ax2)
intensity.plot(ax=ax1)
```
(may need some modifications, but guess my point is to leverage xarray as much as possible).

* Might want to mention in the notebook what `projected_fields` is. I think it's a dictionary of `xr.DataArray`?

### Other notebooks

* A bit unrelated but I noticed that `BoundaryConditions.ipynb` seems to have some broken links, see [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/BoundaryConditions.html#Example-2:-different-boundaries-on-different-edges). Could we fix them in this PR?

shashwat-sh:
Thanks @tylerflex, I've pushed the changes.

> the notebook is still set up to use your local solver, so will need to change to use webapi.

That's strange, I did update those - are you looking at the latest commit? This is what I see: https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/shash/field_proj_2/docs/source/notebooks/FieldProjections.ipynb. Hopefully you'll see the updated version after the current commit anyway.

> You know what's up with this warning?

I think it's a result of computing a grid of angles from reciprocal vectors, which leads to a non-monotonic ordering of angles along a dimension. I'll see if I can reorder the data.

> I'm not a big fan of renormalize_fields seemingly modifying the data contained in the object that calls it. It seems to break the immutability design we have going. I'd rather it returns a copy with the data renormalized 

It does modify the raw fields, so that the other methods can then operate on the most up-to-date fields to return various other quantities, e.g. power, RCS. Yeah we should be able to return an updated copy of `self`, and then the user can use that returned `AbstractFieldProjectionData` object thereafter. Will give that a shot. In any case though, `renormalize_fields` is sort of "harmless" in the sense that one can always revert to original data by re-projecting to the original distance.

> for timing, consider using the ipython magic command %%time instead of time.time(). As far as I remember, using time.time() isnt always the most reliable way to do it.

It's a little inconvenient to get the time as a variable using `%%time` or `%%timeit`.  Looks like `time.perf_counter()` is a bit better for measuring performance than `time.time()`, so I changed to that.

> I feel like maybe this notebook is so long, we should consider breaking it up into a separate Greens Function notebook and have the two notebooks reference each other. Any thoughts?

Yeah it has gotten quite long, but it's also nice to be able to compare the various results in one place, especially as a tutorial / reference like the boundary conditions notebook. Maybe I can add a table of contents at the top with links to help navigate?

> I feel like the fields should be co-located before computing this. Are they co-located?

Yes, these are all projected to the projection grid specified on the focal plane.

tylerflex:
sure, feel free to merge at your convenience, I dont really have any other comments but can take another look later if needed.

PR #73, opened 2022-11-28
Tom/dbr cavity

tomflexcompute:
Hi @momchil-flex could you give this a quick review? There has been some request for examples in 2D and 1D. This notebook simulates a DBR in 1D.


momchil-flex:
- For standalone equations, instead of `<center>$\frac{\Delta f}{f_0} = \frac{4}{\pi} arcsin(\frac{n_1-n_2}{n_1+n_2})$, </center>`, use
```
$$
\frac{\Delta f}{f_0} = \frac{4}{\pi} arcsin(\frac{n_1-n_2}{n_1+n_2}),
$$
```
- Generally we try to format printed output a bit to avoid e.g. too many digits. So this `print(f'The normalized bandwidth of the reflection band is {df}')` could become `print(f'The normalized bandwidth of the reflection band is {df:1.2f}')`.

- `#plot the analytically calculated reflectance at the central frequency as a scatterer`: scatterer -> scatter plot. Or maybe `...central frequency with a star marker`.

- `To obtain the reflectance spectrum of a high-Q DBR cavity, a long simulation run time is required since we need to run the time stepping until the energy in the simulation domain dissipates. Fortunately, for 1D or 2D simulations, this is still an easy task.` This is true, but I think we may also want to take the opportunity to remind the user that when either the Q or the cavity size (or a combination of those) become too large, it is inevitably impractical to simulate until the fields decay, and the user may want to use our ResonanceFinder instead.

- This whole thing
```
Ex = np.array(np.squeeze(sim_data['field'].Ex.sel(f=freq0,method='nearest')))  #extract the on-resonance field data from the field monitor
z = np.array(sim.grid.centers.z)  #extract the grid points in the z direction
Ex = np.delete(Ex,[0,-1])  #make the lens of Ex array and z array consistent

#plot the field distribution
ax1.plot(z, np.abs(Ex))
ax1.set_title('|Ex(x, y)| at $f_0$')
ax1.set_xlabel('z')
```

can be replaced by

```
np.squeeze(sim_data['field'].Ex.sel(f=freq0)).abs.plot(ax=ax1)
ax1.set_title('|Ex(x, y)| at $f_0$')
```

and is a good illustration of the power of using xarray for our data! There's a bit of a learning curve but we should encourage our users to get used to this as it's very powerful.

tomflexcompute:
Thanks for the feedback. The notebook has been revised accordingly. 

When referring to ResonanceFinder, I provided a link to the photonic crystal cavity notebook, which should be using the ResonanceFinder once the new version comes out. In the future, we can also add a link to the ResonanceFinder API reference?

momchil-flex:
Actually I will be adding an updated L3 photonic crystal cavity notebook using the ResonanceFinder which I think will be more relevant. I can change the link when I do that. And yeah I can link the ResonanceFinder itself.

momchil-flex (approved):

tomflexcompute:
Sorry, I'm a little confused. Isn't the cavity notebook been updated to use ResonanceFinder already? I thought it was updated by Casey shortly after the ResonanceFinder was implemented. 

Also do you prefer "squash and merge" or "rebase and merge"?

momchil-flex:
> Sorry, I'm a little confused. Isn't the cavity notebook been updated to use ResonanceFinder already?

You're right, I got confused too, I thought you had provided a link to the photnic crystal band structure notebook, not the cavity one. In any case I am still going to be replacing the L3 cavity notebook with a new one which is based on a paper we recently published. In fact the notebook is already [live](https://docs.flexcompute.com/projects/tidy3d/en/optimized_l3/notebooks/OptimizedL3.html) but not on the develop branch yet.

PR #75, opened 2022-12-02
Added Euler waveguide bend notebook

tomflexcompute:
Hi @dbochkov-flexcompute just wondering if you could give this notebook a quick review? We discussed about formatting example notebooks but don't remember what the conclusion was. Here I ran the black reformatter on this notebook already.


dbochkov-flexcompute (approved):
Looks great! Just one comment: make sure definitions of x and y are consistent with the code (it seems like x should be defined as integral of cos, while y as integral of sin)

I am not sure if these suggestions would be helpful (up to you if you want to incorporate them):
- It seems like scipy has built-in functions to compute values of integrals of this kind https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.fresnel.html
- If I am not mistaken derivative at L_max can be computed analytically as follows: 
`dy/dx(xp) = (dy/dt(L/A)) / (dx/dt(L/A)) = (A*sin((L/A)^2/2)) / (A*cos((L/A)^2/2)) = tan((L/A)^2/2)`. If this is correct maybe it can be used to simplify finding L_max

Regarding formatting, I believe the conclusion was that during releases the black will be used on all notebooks to ensure they have the same style, so if a notebook does not conform to the black style during PR it is not a big deal. But probably it's still good to do to make sure reformatted code doesn't look weird in places.

tomflexcompute:
Thanks for the great feedback! Indeed the x and y definitions in the equations are reversed. I've updated them. 

The scipy built-in function is pretty nice but using it would cause an annoying np.sqrt(np.pi) prefactor floating around since in the Fresnel integral there is a pi. And the analytical derivative is also great but somehow it seems less stable. For the sake of time I'll keep it as it is and find a chance to update them after the 1.8 release. Nonetheless, your time and suggestions are greatly appreciated!

PR #76, opened 2022-12-05
Tom/plasmonic yagi uda antenna

tomflexcompute:
Hi @shashwat-sh could you give this a quick review? It's the same Plasmonic antenna notebook you saw a few weeks before but updated to the new FieldProjection. Thanks!


shashwat-sh (approved):
Looks good overall! Minor comments:

- In the section `Alternative Approach` there's a reference to `Near2FarAngleMonitor` in the text which can be updated to `FieldProjectionAngleMonitor`
- On a related note, it might be useful to link that to our docs, e.g. [FieldProjectionAngleMonitor](https://docs.flexcompute.com/projects/tidy3d/en/latest/_autosummary/tidy3d.FieldProjectionAngleMonitor.html) (though that link right now won't work until we build and publish the latest docs, which will happen when 1.8 is released). Would be nice to do the same for FluxMonitor and PointDipole, e.g. in the text above cell [5] or wherever they are first mentioned. You can get links to these from https://docs.flexcompute.com/projects/tidy3d/en/latest/api.html
- `Near2Far` is also mentioned in some comments, e.g. `# create a Near2FarAngleMonitor...` in cell [5]

tomflexcompute:
Thanks for the review. Great catch! Those have been updated now. 

Actually earlier I wanted to add the links and realized the problem that you mentioned. We can merge it as it is now and easily circle back to update this shortly after the 1.8 release. Do we have a specific date for 1.8 release already?

shashwat-sh:
Hmm might be a good question for @tylerflex to see if we should add the links now even if they don't exist, or re-build the docs later. The only concern is that releasing and then re-building the docs might be a bit of an inconvenience.

Other than that, looks good to me, thanks! Not sure about the specific release date, but probably by the end of the week?

shashwat-sh:
And I guess you can just merge this PR anyway and make the link updates separately.

tomflexcompute:
Thanks. Either way would be fine. I think the URL for the updated API reference is clear so the link won't break after release. I can add them now if that's preferred.

tomflexcompute:
API reference links are added. Would you mind taking a quick last look @shashwat-sh before I merge?

shashwat-sh:
Looks good to me!

PR #77, opened 2022-12-05
Adding mode sorting and broadband source features

dbochkov-flexcompute:
Adding some basic demonstration of broadband sources
- Added mode sorting into "Mode Solver" tutorial 
- Added broadband basic usage into "Mode sources and mode monitors" tutorial (btw, in the current version flux plots don't look right for some reason https://docs.simulation.cloud/projects/tidy3d/en/latest/notebooks/Modal_sources_monitors.html#Mode-monitors)
- Switched to broadband source in "Ring resonator" example


tomflexcompute:
Thanks for the update. Looks great overall. I think the broadband source is very easy to use and will benefit a lot of simulation cases. 

I did some typo checks and found a few, which I listed below. Most typos probably have been there for a while and are not from your update this time, but it would be great if you can correct them along the way. Also some of my questions and suggestions are listed below.

ModeSolver Notebook:
•	In the markdown above cell [26], I think “follow” should be “follows”
•	In the markdown above cell [32], ‘specificall’ seems to be incomplete? 
•	In the markdown above cell [87], ‘egeinvalue’ should be spelled “eigenvalue”.
•	The ModeSolver gives warnings that field does not decay at the plane boundaries. I wonder if those modes are physical modes (radiative modes?) or some numerical artifacts? If some modes are unphysical and purely numerical solutions, should we consider getting rid of them?

Mode sources and mode monitors notebook
•	For the transmission power larger than 1, is it because the injected power is too low outside of the frequency window?
•	In the markdown above cell [5], “for the fields do decay…” should be “for the fields to decay..”.
•	In the markdown above cell [19], “wveguide” should be spelled “waveguide”.
•	In the markdown above cell [11], “the radiated power in to forward- and…” sounds strange. Do you mean to say “the radiated power in the forward- and…”?
•	In the markdown above cell [13], “optiones” should be spelled “options”.
•	In the markdown above cell [23]. “juction” should be spelled “junction”.

Ring resonator notebook
•	Cell [16] and [17] give a very long console output. It’s not crucial but it would improve the readability of the notebook if these can be hidden somehow.
•	Above [3], “walength” should be spelled “wavelength”.

tomflexcompute (approved):

dbochkov-flexcompute:
@tomflexcompute, thanks a lot for catchings all of those, they are fixed now

> The ModeSolver gives warnings that field does not decay at the plane boundaries. I wonder if those modes are physical modes (radiative modes?) or some numerical artifacts? If some modes are unphysical and purely numerical solutions, should we consider getting rid of them?


I believe those are radiative modes (@momchil-flex, please correct me if I am wrong)




> In the markdown above cell [32], ‘specificall’ seems to be incomplete?

Actually, I'm not sure what was supposed to go in there. @momchil-flex, maybe you would have a better idea? here's the context:
> Screenshot




> For the transmission power larger than 1, is it because the injected power is too low outside of the frequency window? 

I assume you are talking about this plot? and I guess here we have transmission power =  injected power, since it's a straight waveguide. I think we can explain this in the following way. When we inject waveguide modes we actually have two normalizations at each frequency: first, by the flux of the field distribution being injected, so that it has flux = 1, and, second, by the signal amplitude at the given frequency. I think the injected power greater than 1 is primarily caused by the first normalization. That is, away from frequency points where we do calculate modes (either `freq0` or  `+/- 1.5 * fwidth`) our flux calculation does not match what field distribution would actually be propagating, either overestimating or underestimating its flux. Thus, causing either increase or descrease in injected power. (@momchil-flex, again please correct me if I am not quite right here)
> Screenshot from 2022-12-05 14-50-47

momchil-flex:
> @tomflexcompute, thanks a lot for catchings all of those, they are fixed now
> 
> > The ModeSolver gives warnings that field does not decay at the plane boundaries. I wonder if those modes are physical modes (radiative modes?) or some numerical artifacts? If some modes are unphysical and purely numerical solutions, should we consider getting rid of them?
> 
> I believe those are radiative modes (@momchil-flex, please correct me if I am wrong)

They are actually unphysical, because by default the mode solver uses PEC boundaries at the plane boundaries. In the actual simulation, there is no such PEC box surrounding the mode propagation. The mode solver actually also supports PML truncation to the plane, in which case radiative modes could be computed. However, the PMLs also introduce many spurious modes which are hard to filter out and generally make it much harder to use those boundaries than the default PEC ones. Practically speaking, only modes that decay at the plane edges should be used as mode sources and in mode monitors. However, it's a bit hard to decide on an automatic truncation factor - sometimes you may have a bit of field at the boundaries and still want to inject the mode (maybe you have no choice...) That said I do agree that it would be nice to avoid printing so many warnings, I'm just not sure whether automatically filtering the modes can be done in a safe way. What does Lumerical do? I believe you'd also get a lot of spurious modes and they also use PEC boundaries by default? And it's up to the user to select the right one? We *could* just remove the warning, or any other suggestions?

> > In the markdown above cell [32], ‘specificall’ seems to be incomplete?
> 
> Actually, I'm not sure what was supposed to go in there. @momchil-flex, maybe you would have a better idea?

Ehm I don't remember either. :D Let's just remove...

> > For the transmission power larger than 1, is it because the injected power is too low outside of the frequency window?
> 
> I assume you are talking about this plot? and I guess here we have transmission power = injected power, since it's a straight waveguide. I think we can explain this in the following way. When we inject waveguide modes we actually have two normalizations at each frequency: first, by the flux of the field distribution being injected, so that it has flux = 1, and, second, by the signal amplitude at the given frequency. I think the injected power greater than 1 is primarily caused by the first normalization. That is, away from frequency points where we do calculate modes (either `freq0` or `+/- 1.5 * fwidth`) our flux calculation does not match what field distribution would actually be propagating, either overestimating or underestimating its flux. Thus, causing either increase or descrease in injected power. (@momchil-flex, again please correct me if I am not quite right here)

Yes, I believe your explanation is correct. And I have seen power > 1 away from the central frequency even in the case of single-frequency source so I am not particularly worried. I just hope broadband sources don't go completely out of control outside of the +/- 1.5 * fwidth region, but so far there's no indications that that's the case.

Some further comments.

Mode solver notebook:
- I'm afraid that the current illustration of the sorting is largely (entirely?) applied to the non-physical modes. Apart from the fields not decaying at the boundaries, no mode with effective index < 1 should be considered physical. I think that unfortunately a bit more effort will be needed here to pick a waveguide where *physical* modes cross. This could be done by adding a substrate as initially planned, although it can probably also happen in a single waveguide if the width and height are well-tuned.
- In your many-panel plots, the waveguide orientation is transposed (long side is along the y-axis). You need to add a `.T` here `ax[j, i].imshow(abs(mode_data.Ex.isel(mode_index=j, f=i, drop=True).squeeze()), cmap="magma")`, or use the approach used earlier in the notebook, `abs(mode_data.Ex.isel(mode_index=0, f=f0_ind)).plot(x='x', y='z', ax=ax1, cmap="magma")`. But maybe there was a good reason you preferred to use `imshow` directly, in which case yeah just add a transpose.

Mode sources and monitors notebook:
- "Duplicate previously defined simulation and updated their source lists." -> "Duplicate previously defined simulation and update the source lists."
- In the last plot, `ax[mode_index].set_title("Mode decomposition (+ propagating)")` -> `ax[mode_index].set_title(f"Mode {mode_index}, direction='+'")`

Ring resonator notebook:
- This discussion keeps getting stale:
    "With a center wavelength of 500 nm and 10 nm resolution, this is a challenging FDTD problem because of the large simulation size. The simulation contains 2 million grid points to model the entire domain and 290,000 time steps to capture the resonance of the ring.
    With Tidy3D, we perform the simulation in just a few minutes."

    Let's just rephrase it to something like
    "With a center wavelength of 500 nm and 10 nm resolution, this is a challenging FDTD problem because of the large simulation size and the large number of time steps needed to capture the resonance of the ring. However, it can be easily handled with Tidy3D."
- "we lower the shut off conditions" -> "we lower the shutoff factor" 
- Do you know why the transmission seems to dip to 0.99 in the low-frequency side of the last plot?

dbochkov-flexcompute:
> Yes, I believe your explanation is correct. And I have seen power > 1 away from the central frequency even in the case of single-frequency source so I am not particularly worried. I just hope broadband sources don't go completely out of control outside of the +/- 1.5 * fwidth region, but so far there's no indications that that's the case.

yeah, I was worried about too, but when I checked `+/- 8 * fwidth` span I didn't see anything weird (the wider gray area in the figure is `+/- 4 * fwidth`)
Screenshot

> I'm afraid that the current illustration of the sorting is largely (entirely?) applied to the non-physical modes. Apart from the fields not decaying at the boundaries, no mode with effective index < 1 should be considered physical. I think that unfortunately a bit more effort will be needed here to pick a waveguide where physical modes cross. This could be done by adding a substrate as initially planned, although it can probably also happen in a single waveguide if the width and height are well-tuned.

Sounds good, on it!

> But maybe there was a good reason you preferred to use imshow directly

all axes, labels, and titles clutter this figure very much, with `imshow` it is less to clean

> Do you know why the transmission seems to dip to 0.99 in the low-frequency side of the last plot?

I haven't investigated that, it is consistent with the previous version of the notebook, so I assumed this was expected

momchil-flex:
> I haven't investigated that, it is consistent with the previous version of the notebook, so I assumed this was expected

I see, if I have to guess the mode becomes more delocalized and gets scattered by the ring even away from the ring resonance frequencies. I think it's fine.

tomflexcompute:
> They are actually unphysical, because by default the mode solver uses PEC boundaries at the plane boundaries. In the actual simulation, there is no such PEC box surrounding the mode propagation. The mode solver actually also supports PML truncation to the plane, in which case radiative modes could be computed. However, the PMLs also introduce many spurious modes which are hard to filter out and generally make it much harder to use those boundaries than the default PEC ones. Practically speaking, only modes that decay at the plane edges should be used as mode sources and in mode monitors. However, it's a bit hard to decide on an automatic truncation factor - sometimes you may have a bit of field at the boundaries and still want to inject the mode (maybe you have no choice...) That said I do agree that it would be nice to avoid printing so many warnings, I'm just not sure whether automatically filtering the modes can be done in a safe way. What does Lumerical do? I believe you'd also get a lot of spurious modes and they also use PEC boundaries by default? And it's up to the user to select the right one? We _could_ just remove the warning, or any other suggestions?
> 
I believe spurious modes are inevitable and hard to filter universally in all mode solvers. The users do have to decide which modes they want to use and include. Maybe we can set up a notebook to demonstrate how to filter out spurious modes but certainly that can only be done on a case by case basis. Simply setting the effective index in a reasonable range and making sure the field decays at the boundaries should do most of the work.

By the way, do we still want to demonstrate the broadband source in other notebooks that are currently using only the central frequency?

dbochkov-flexcompute:
Applied all suggestions and switched mode tracking section to a layered waveguide https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/402e09421b641a9b1e0ae37f43e70bc8b35ceba0/docs/source/notebooks/ModeSolver.ipynb

PR #82, opened 2022-12-13
Added waveguide size converter notebook and the schematic files

tomflexcompute:
Hi @caseyflex could you give this notebook a quick review when you have time? No hurry if you are busy with something more urgent. We can also merge this after the 1.8 release. Thanks!


caseyflex (approved):
Looks great! One comment is that for the Luneburg lens, another picture might help convey that you are using an effective index model rather than doing a full 3D simulation of this structure.

tomflexcompute:
Thanks @caseyflex for reviewing. @momchil-flex this branch is ready to be merged. You can do so when you think it's a good time. I think it should be fairly safe but to avoid further complications, maybe it's a good idea to do so after 1.8 release.

PR #88, opened 2022-12-27
Added THz demultiplexer notebook and schematic image

tomflexcompute:
Hi @momchil-flex could you give it a quick review?

A few customers requested models related to ring resonators. We have a 2D ring resonator example but seems like a 3D example would be beneficial to have. I have sent this notebook to a few customers and they found it helpful so I figure we can make it public. This notebook demonstrates a THz ring resonator device (very similar to that in near-IR really). I deliberately chose THz to broaden the scope of our current example libraries. 

The only downside that I see is this model requires a reasonably long run time and about 6 flex units, even though the model itself is not that big.


momchil-flex:
Overall looks great!

Small comments:
- Is the cylinder bend in the drop port how it's defined in the paper? I would imagine that this type of bend is somewhat lossy, but actually the drop port transmission seems pretty good! Probably the radius is large enough...
- Why do you have `ax.set_aspect("auto")` when plotting the fields? The default `"equal"` aspect ensures that the structure shapes are preserved. In your plots, the ring looks like an ellipse.

Regarding the cost: we've been thinking about adding a subsection of examples that have a higher than negligible cost. It seems to me that we may already have some other examples that may also go there, e.g. the Yagi-Uda antenna. So should we create that subsection? What should we call it? Should we also add a sentence in the beginning of each example warning about the cost? How should we phrase that? @xin-flex , thoughts?

tomflexcompute:
Thanks for reviewing. 

- Yes, the radius is sufficiently large, comparable to the radius of the ring, so the loss is quite small. 

- Good point. Actually I don't remember why I set the aspect ratio to auto in this case. It has been updated to the default.

tomflexcompute:
A comment about using 6 flex units is added in the markdown. 

Since we are adding depreciating warnings now, shall we merge this later on? This model uses pml on all boundaries and they are explicitly set so no warning should occur.

xin-flex (commented):
minor edit: change "flex units" to "FlexUnit"

momchil-flex:
I'll just merge this now. Then when I rerun all notebooks with the actual flex unit printed, we can see if we should add the same comment to some of the other ones.

momchil-flex:
I think the comment should be on top of everything else though - right under or even above the "run this notebook in binder". I think we should just have a standardized comment we put in all notebooks that take larger than x credits (x = 1?)

tomflexcompute:
I agree. I can change it and other notebooks as well if we can finalize the format.

PR #89, opened 2022-12-27
Fixed header level

tomflexcompute:
Just realized that on the website under '[Case studies](https://docs.flexcompute.com/projects/tidy3d/en/latest/examples.html#case-studies)', there is an entry called 'Postprocessing'. This is due to one header of the Plasmonic Yagi Uda antenna notebook being incorrect. It's fixed here so next time when the website is rebuilt, the issue should go away.


momchil-flex:
Thanks, good catch. I'll merge this into the readthedocs branch too so that our "latest" build updates.

PR #87, opened 2022-12-16
Replace gdspy with gdstk for GDSII import

lucas-flexcompute:
Update the notebooks that use `gdspy` to use `gdstk`, as well as the documentation.

There are also a few other minor improvements throughout the notebooks aiming at simplification.

PR #92, opened 2023-01-05
Added notebook and associated files

tomflexcompute:
This is a relatively long notebook so better have more reviewers to make sure we catch and fix all potential issues. Thanks!

@xin-flex could be a good demo model to use in presales meetings for customers who are interested in doing large PIC simulations.


tylerflex (changes requested):
Hey Tom, this is quite an impressive effort! Thanks a lot for the notebook. Overall looks great, just a few relatively minor comments from me:

1. Intro
	1.  hyperlink to ModeSolver in docs.
	2. “aim to model” -> “model” (seems a bit weird to say “aim” in this context).
	3. We will soon switch from `gdspy` to `gdstk` internally and in notebooks, maybe we should do that here too?
	4. What warnings are being suppressed here? Are they truly unnecessary? It might be helpful to explain why it is ok to suppress the warnings this time so the user doesn’t think they should ignore them generally.
2. Mode Indices Calculation
	1. "on the local computer” -> “on a local computer”
	2. `neff[i, :]` can actually just be `neff[i]`  using numpy broadcasting rules. Slightly cleaner to avoid `:`. [More details](https://numpy.org/doc/stable/user/basics.indexing.html)
3. TE0 to TEX conversion
	1. Where do you come up with the geometric values, eg in `sim = make_sim(0.4, 0.835, 0.2, 15.5, “TE”)`? Directly from the paper? I wonder if it would be somewhat more clear if you constructed a dictionary / table to store all of these values in one place at the beginning of the notebook, which can then be grabbed later when you are making these sims. (Optional suggestion)

tomflexcompute:
Thanks for the great suggestions @tylerflex. All are incorporated besides the `gdspy` change. How soon will the switch to `gdstk` be complete? Maybe it's a question for @lucas-flexcompute.  Does `gdstk` have a similar python API as `gdspy`? If so, would be an easy fix.

tylerflex (approved):
I'll let @lucas-flexcompute answer regarding whether it's worth changing and how much additional effort it is. Otherwise, notebook looks good from my perspective! thanks

tomflexcompute:
Thanks @tylerflex and @lucas-flexcompute for reviewing. 

@lucas-flexcompute Would you mind spending some time converting `gdspy` to `gdstk` on this notebook alone with other notebooks? Actually is the [GDSII import](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/GDS_import.html) notebook the only one we have that uses gdspy?

lucas-flexcompute:
There are only 3 notebooks making use of `gdspy` at this point (see https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/87) and in tidy3d it is only used for GDSII import.

You can make the geometry in a more straightforward way like so:
```python
    y = By + (w_access + w_bus) / 2 + gap
    access_wg = gdstk.RobustPath((-3 * l_couple, y), w_access, simple_path=True, layer=1, datatype=0)
    access_wg.segment((-l_couple / 2 - Bx, y))
    access_wg.segment((-l_couple / 2, y), offset=lambda u: (np.cos(u * np.pi) - 1) * By / 2)
    access_wg.segment((l_couple / 2, y))
    access_wg.segment((l_couple / 2 + Bx, y), offset=lambda u: (np.cos((1 - u) * np.pi) - 1) * By / 2)
    access_wg.segment((3 * l_couple, y))
    
    # calculate coordinates for the bus waveguide
    bus_wg = gdstk.FlexPath([(-3 * l_couple, 0), (3 * l_couple, 0)], w_bus, layer=1, datatype=1)

    # define a cell
    cell = gdstk.Cell("directional_coupler")
    cell.add(access_wg)
    cell.add(bus_wg)
```
Here there is no need for functions `bend_path` or `access_wg_path` (and it should work almost the same way with `gdspy` too, with minor changes).

I've also just noticed that you can make the simulation call much simpler as well:
```python
sim = make_sim(pol="TE", **design_params['TE3'])
```

tomflexcompute:
Thanks @lucas-flexcompute . I pulled `lucas/gdstk` branch and switched to gdstk in the notebook. However when I ran a job, it errored. From the log it seems to be related to using `AutoGrid`. Have you seen this issue or know that's wrong?

xin-flex (approved):
This example notebook is very impressive! Thanks Tom for the great work! I have nothing to add to it. It would be very interesting to showcase to potential PIC users!

lucas-flexcompute:
> Thanks @lucas-flexcompute . I pulled `lucas/gdstk` branch and switched to gdstk in the notebook. However when I ran a job, it errored. From the log it seems to be related to using `AutoGrid`. Have you seen this issue or know that's wrong?

Hi @tomflexcompute, I believe it was because my branch was lagging behind develop. I've just merged the changes introduces after I opened the PR, so it should work. (I can't test right now, but I can do it tomorrow if the issue remains).

tomflexcompute:
> 

Thanks @lucas-flexcompute . Now it works very well. And thanks again for the simplifications. These are really great ideas. Would you mind taking a quick last look at the current version of the notebook? If everything looks fine I think it's ready to be merged.

lucas-flexcompute:
Looks great to me!

lucas-flexcompute (approved):

PR #102, opened 2023-01-17
Added the gradient metasurface reflector notebook and associated schematic

tomflexcompute:
Hi @dbochkov-flexcompute and @caseyflex . Could you give this gradient metasurface reflector notebook a quick review? Thanks in advance!


dbochkov-flexcompute (approved):
Looks great! Here're my comments:
- it looks like arrays `ldas` and `freqs` are not really used for anything
- is it worth mentioning explicitly that glass and spacer material will be modeled as simple dieletric materials, while gold will be modeled as a dispersive material?
- "According to the reference, we design a super cell that consists of 10 antennas of 5 different lengths" - do you mean that you are repeating the design from the paper?

caseyflex (commented):
Looks great overall!

In the result analysis (cell 19), I find the plot a bit unclear. Why are two of the diffraction orders at the same angle?

tomflexcompute:
Hi @dbochkov-flexcompute . Thanks for the great suggestions. I've revised the notebook accordingly. 

Hi @caseyflex . That's a great point. I agree that it is a little confusing. Currently the diffraction angles from `DiffractionMonitor` doesn't distinguish signs. Therefore, +1 diffraction order and -1 diffraction order have the same angle under normal incidence. I think it would be better to label the angles with signs but I guess that's up to personal preference. We can discuss with others about it later. For now, I added a comment in the markdown to explain this. 

Thanks both for the quick review! I'll leave it here for today in case you want to take another look. If there are no further comments I'll merge it tonight.

tomflexcompute:
Actually I revised the plotting a little bit to make the diffraction angles more clearly. This is also consistent with the GUI way of plotting DiffractionMonitor results.

caseyflex (approved):

PR #103, opened 2023-01-24
Added photonic crystal filter notebook

tomflexcompute:
Hi @weiliangjin2021 could you give this photonic crystal waveguide mode filter a quick review? Thanks!


weiliangjin2021:
Very good! A few comments:
- x/y/z coordinates are mentioned below Cell 4, but not defined in the schematic. Wonder if it can be added in the figure.
- When defining the waveguide, it appears to me that you can just define a single waveguide of length td.inf along x-direction. The part inside the PhC will be overridden by the holes if the holes are added the last.

weiliangjin2021 (approved):

tomflexcompute:
Thanks for the helpful comments. The notebook has been revised, in case you want to take another look. Will merge it later today.

tomflexcompute:
@tylerflex this PR is ready to merge into develop, but since we opened a new branch for pre release, I wonder if we should merge this into the pre release branch instead?

tylerflex:
does it work with 1.8.2? if so, I can just push it to readthedocs and make it public now?

tomflexcompute:
> does it work with 1.8.2? if so, I can just push it to readthedocs and make it public now?

Yes, it does not depend on any new features so in principle can be published as it is now.

PR #107, opened 2023-01-27
Move anisotropic medium to dispersive list

weiliangjin2021:


tylerflex:
Hm, maybe the distinction between "Dispersive Media" and "Non-Dispersive Media" should be rethought, because it seems to me that AnisotropicMedium doesn't really fit into either because it can be either dispersive or non-dispersive. Maybe we should include a new category that covers a medium that is a combination of other media. For now, perhaps we call this "anisotropic media"?

## Non-Dispersive Media
* Medium
* PEC

## Dispersive Media
* Lorentz
...

## Anisotropic Media
* AnisotropicMedia
* (Medium2D)

What do you think?

weiliangjin2021:
Where does CustomMedium fit in that case? Perhaps in the catogary "anisotropic medium", but add a note that it's dispersiveless?

tylerflex:
Hm, maybe either in anisotropic media or we could rename that section "Other Mediums"?

tylerflex:
sounds good to me

weiliangjin2021:
sounds good to me as well

tylerflex:
implemented in #120

PR #104, opened 2023-01-25
add pre-release adjoint tutorials

tylerflex:

PR #109, opened 2023-02-01
Tyler/clean adjoint pre 1.9

tylerflex:


tylerflex:
merging instead of #104 because it was easier than dealing with the submodule merge conflicts

PR #115, opened 2023-02-10
Self-intersecting polyslab tutorials

weiliangjin2021:


tomflexcompute (changes requested):
Thanks @weiliangjin2021 . Looks great. A few comments are listed below for your reference. Also, I think you need to update the examples.rst file as well?

1.	In the current convention, only the first letter in the notebook title is capitalized. Since `PolySlab` is a class name, I think it’s fine to make it either “Self-intersecting PolySlab” or ” Self-intersecting polyslab”.
2.	pydantic is imported but never used? Does it have to be imported?
3.	Above cell[6], “an ComplexPolySlab” should be “a ComplexPolySlab”.
4.	In the GDS import part, the layer is first saved to a .gds file, then the file is read. Is this necessary? The cell can be directly used in `from_gds` after being defined.

weiliangjin2021:
Thanks, Tom. Please see the revised notebook.

tomflexcompute (approved):
Looks good! Thanks.

PR #116, opened 2023-02-10
adjoint notebooks for 1.9.0rc2

tylerflex:
Note that this also updates all hyperlinks for rc2


tomflexcompute (commented):
The notebooks look great! The third tutorial using `CustomMedium` works fantastically. Just have a few questions:

1. In the second notebook, there are a lot of overlapping warnings. It's also mentioned in the markdown that "One must therefore be careful about overlapping". Just wondering if these overlapping will cause any issue? Maybe more explanation in the markdown regarding this can be helpful especially for new users.

2. For practical purposes, I think binarized optimization will be more common and desirable. If one were to do a binarized optimization, would it require a different approach, or you can take the result from the continuous optimization and manually binarize the optimized eps according to some threshold?

tylerflex:
Hey @tomflexcompute , thanks for reviewing.

> In the second notebook, there are a lot of overlapping warnings. It's also mentioned in the markdown that "One must therefore be careful about overlapping". Just wondering if these overlapping will cause any issue? Maybe more explanation in the markdown regarding this can be helpful especially for new users.

So, the overlapping will cause gradient error issues if the geometry center or size is changing. To help explain this, the updated notebook contains a long discussion about the cause of errors with overlapping structures and how to avoid them. The notebook is also modified a bit to handle this case, essentially we fix the center positions of the slabs and don't track gradients through them, which would cause the errors. 

Unfortunately, the many log warnings are just part of the validator and I think should be left in just in case the user doesn't realize the issue. 

What do you thnik? 

> For practical purposes, I think binarized optimization will be more common and desirable. If one were to do a binarized optimization, would it require a different approach, or you can take the result from the continuous optimization and manually binarize the optimized eps according to some threshold?

I agree and am working on a tutorial for the next release that does some binarization and smoothening of the features so they are larger and more fabric able. It's a bit of an art and is taking a while for me to put together these notebooks, so my preference would be to release it as is for now with the performance improvements, but as soon as the new tutorials are ready, we can replace this example or do some binarization / filtering. It's high on my priority list and I think very important for practical purpoes.

tomflexcompute:
> Hey @tomflexcompute , thanks for reviewing.
> 
> > In the second notebook, there are a lot of overlapping warnings. It's also mentioned in the markdown that "One must therefore be careful about overlapping". Just wondering if these overlapping will cause any issue? Maybe more explanation in the markdown regarding this can be helpful especially for new users.
> 
> So, the overlapping will cause gradient error issues if the geometry center or size is changing. To help explain this, the updated notebook contains a long discussion about the cause of errors with overlapping structures and how to avoid them. The notebook is also modified a bit to handle this case, essentially we fix the center positions of the slabs and don't track gradients through them, which would cause the errors.
> 
> Unfortunately, the many log warnings are just part of the validator and I think should be left in just in case the user doesn't realize the issue.
> 
> What do you thnik?
> 
> > For practical purposes, I think binarized optimization will be more common and desirable. If one were to do a binarized optimization, would it require a different approach, or you can take the result from the continuous optimization and manually binarize the optimized eps according to some threshold?
> 
> I agree and am working on a tutorial for the next release that does some binarization and smoothening of the features so they are larger and more fabric able. It's a bit of an art and is taking a while for me to put together these notebooks, so my preference would be to release it as is for now with the performance improvements, but as soon as the new tutorials are ready, we can replace this example or do some binarization / filtering. It's high on my priority list and I think very important for practical purpoes.

@tylerflex yeah I think that makes sense. 

Did you modify the second notebook? Which paragraph is the discussion you are referring to?

docs/source/notebooks/AdjointPlugin_2_GradientChecking.ipynb
```diff
@@ -219,21 +191,16 @@
...
     "To compute the gradient of our transmission with respect to each of the slab thicknesses and permittivities, we need to repeat this step for each of the values. Luckily, since TMM is very fast, we can compute these quantities quite quickly compared to if we were using FDTD.\n",
     "\n",
-    "Here we write a function to return this gradient."
+    "> Important note: We assume in our TMM numerical gradient that when the slabs are touching (`spc=0`) and a slab thickness is modified, that the thicknesses of the neighboring slabs adjust to accomidate this change. For example, if slab `i` increases by `dt`, slab `i-1` and `i+1` each decrease by `dt/2`. We also account for this in our FDTD set up by keeping the centers of all boxes constant and not tracking the gradient through these quantities. The reason this is required is that `tidy3d.plugins.adjoint` does not recognize the space between touching `JaxBox` objects as a single interface and will instead \"double count\" the gradient contribution of the interface if they are placed right next to each other. One must therefore be careful about overlapping or touching two `JaxBox` or other geometries when computing gradients.\n",
```
tylerflex, line 194: @tomflexcompute see this note

tomflexcompute (approved):
Reads well. Thanks! Feel free to merge as it is.

PR #119, opened 2023-02-14
show inherited members by default

tylerflex:


momchil-flex (approved):
I guess let's get this in and see what happens :)

PR #122, opened 2023-02-15
Added graphene example

tomflexcompute:
Added the graphene example. As discussed, the `sim = sim.volumetric_equivalent()` lines at cell[7] and [12] need to be removed after the solver is rebuilt and the notebook needs to be tested again.


caseyflex (changes requested):
Hi Tom,

Mostly looks great. A couple comments:
1. I'm not sure we need to mention in the beginning the volumetric equivalent. Since this is operating on a single plane of the Yee grid, it's equivalent to a surface conductivity model.
2. I think you have the last two figures backwards. The Medium2D.plot_sigma plots the surface conductivity of the actual medium, which has already been fitted using Pade approximants. On the other hand, Graphene.numerical_conductivity is actually the "analytical" result. It is called numerical because it involves a numerical integration, but this is actually the ground truth while Medium2D.plot_sigma is the approximation.

momchil-flex:
Sorry, could you drop the last commit, as this PR will not go into 1.9.0rc2.

tomflexcompute:
> Sorry, could you drop the last commit, as this PR will not go into 1.9.0rc2.

Sure no problem. We can examine it and merge this at a later stage. Will there be a 1.9.0rc3?

momchil-flex:
To be decided on Monday.

PR #123, opened 2023-02-17
Inclusion of Edge Coupler notebook.

e-g-melo:


tomflexcompute (changes requested):
Thanks @e-g-melo for the great example. I think this example will be referenced by many users in the PIC space as a few of them have explicitly requested such an example model. I just have some minor comments:

•	I think for notebooks that cost above 1 FlexUnit, we just say it costs above 1 FlexUnit regardless of how much it costs exactly. This is to avoid manually changing the number in the future if we change how we charge the simulations. 
•	Just wondering what kind of warnings would you get if the logging level is not set to “error”? Are these warnings important for users to know?
•	To prevent output like “<AxesSubplot: title={'center': 'cross section at x=27.32'}, xlabel='y', ylabel='z'>” being printed to the console, you can simply add a “;” at the last line such as “ax2.set_ylim(-2, 2);”.
•	Just curious, does the shutoff=1e-7 give a different result compared to the default 1e-5?
•	If it doesn’t take too much time, would be pretty cool to demonstrate how to numerically compute the overlap integral of the fiber mode and the taper tip mode. 
•	Since the tapers are pretty long especially in the sweep, one can consider plotting the field distributions using ‘auto’ aspect ratio. I’m not sure if it looks better but feel free to give it a try.
•	The results are very nice. I’m also curious about why quadratic taper performs better at shorter length but worse at longer length. 
•	A few typos are listed below:
“FlexUnits” instead of “Flex Units”
“launched” instead of “lauched”
“around” instead of “arround”
“the quadratic taper profile has” instead of “have”

tomflexcompute:
Also I'm wondering which branch it should be merged into. Should it be the pre1.9 branch, develop, or readthedoc? I'm also a little confused currently. Perhaps @tylerflex can clarify.

tylerflex:
Hi @e-g-melo Thanks a lot for this contribution! the notebook is very well written and I think will be a very useful addition to our case studies.  

I just had a few minor suggestions about how to improve the flow and readability of the python / plotting portions mainly.

One other general comment is that (going forward), rather than merging changes from your own fork of tidy3d-docs, you may want to make your own branches (eg. `emerson/edge_coupler`) from the main tidy3d-docs branch. That way it's a bit easier for other people to check out your branch from the central repository without needing to clone your fork.

Here are my more specific comments for you to take a look and fix when you get time, thanks again!

* Intro paragraph:
	* remove “Basically, there are two main categories…” and just use “There are two main categories…”
* Cell [1]
	* do we really want to set the logging level to “error”? We should avoid this whenever possible as the warnings can be important to consider.
	* Seems like it’s using front end 1.8.2, we should probably upgrade it to 1.8.4 or pre/1.9.0.
* Cell [10]
	* Might be nice to  have the keys of the `sims` dict be the corresponding tap names. `sims = dict(zip(tap_names, sim_tap))` or just modify `get_simulations` to return a dict.
* Cell [14]
	* let’s throw in a `plt.show()` to suppress the text output of `(-2, 2)`
* Cell [15]
	* can probably be simplified a bit like
```py
plot_colors = ('black', 'red', 'blue')
for data, color, label in zip(trans_data, plot_colors, tap_names):
    ax1.plot(
        wl_range,
        data,
        color=color,
        linestyle="solid",
        linewidth=1.0,
        label=label,
    )
```
* The line
`trans_data.append(10 * np.log10(np.abs(coeffs_f.sel(mode_index=0)) ** 2))`
Is hard to read, can we split this up into a few lines? For example, something like
```py
power_0 = np.abs(coeffs_f.sel(mode_index=0)) ** 2)
power_0_db = 10 * np.log10(power_0)
```
(Adjust the variable names as you see fit).
* Inverted Taper Length sweep
	* “ranged” -> “varied”
* Cell [18], 
	* let’s use an f-string (eg. `f”sim_{len(taper_sweep) - 1}_0”` instead of `str.format`. 
* Cell [19]
	* The lines
```py
coup_eff_tl[i, 0] = trans_data[i][len(trans_data[i]) // 2]
```
Could be simplified a bit.
		*  First, `trans_data` is still an `xr.DataArrray` so can we `sel` the central frequency instead of indexing directly at the halfway point to be more explicit?
		* Second, maybe we can do this in a for loop over `i`. 
* Cell [20]
	* Can we first process the x,y data before plotting to make it easier to read?
```py
taper_lengths = [taper_l] + taper_sweep[:]
```
And then loop over the `coup_eff_tl` , `colors`, dot style, and labels to plot, similar to suggestion from cell [15]
* Cell [21] 
	* Let’s use f-strings, as in cell [18]’s comments.
* Cell [20] Same comment as cell [20] about simplifying / generalizing plotting.

tylerflex:
@tomflexcompute I think we can merge this into develop for now. and if we want, I could check out the file into pre/1.9.0 so it's available in both but for now, let's stick with develop.

tylerflex:
closing this because it seems it's taken care of in #130

PR #124, opened 2023-02-17
Tom/case study subsections

tomflexcompute:
Subsections are added. 

I noticed one notebook (PhotonicCrystalWaveguidePolarizationFilter.ipynb) that was merged to pre 1.9 a few weeks ago is gone. Not sure if I accidentally reverted it or something. I added it back here. The notebook has been reviewed so should be good as it is but feel free to take a look at it if you like.


docs/source/examples.rst
```diff
@@ -93,31 +93,53 @@ Pre and Post Processing
...
+    notebooks/GradientMetasurfaceReflector
+    notebooks/Metalens
+    notebooks/HighQ_Si
+    notebooks/HighQ_Ge
+    notebooks/GratingEfficiency
+    notebooks/BiosensorGrating.ipynb
+    notebooks/DistributedBraggReflectorCavity
+
+Photonic Crystals
+~~~~~~~~~~~
```
tylerflex, line 129: I think this will only work properly if the underline is the same length as the title (

```
Photonic Crystals
~~~~~~~~~~~~~~~~~
```
same for the other subsections

tylerflex (changes requested):
Subsections look great, thanks! Just one comment on how rst formats subsections to check out.

docs/source/examples.rst
tomflexcompute: Thanks for the reminder! I didn't notice that requirement. Updated in the current version.

tylerflex (approved):
looks good to me, just compiled locally and works. Feel free to merge once @momchil-flex takes a look

tylerflex:
It would be nice to figure out how to do subsections on the left hand side table of contents, but I haven't figured out how.
image

momchil-flex:
> It would be nice to figure out how to do subsections on the left hand side table of contents, but I haven't figured out how. 

Maybe we can investigate a different page layout for Tidy3D 2.0?

Otherwise looks good. Not sure what happened to the ``PhotonicCrystalWaveguidePolarizationFilter.ipynb``. I think we can merge this in, and then retag/rebuild the 1.9.0rc2 docs. However, if we're doing that, could you make sure links in that notebook point to en/1.9.0rc2 version of the docs?

@tylerflex we really need to figure out how to automate this... I think there must be a way for sphinx to handle it automatically..

tylerflex:
Rather than re-build / re-tag 1.9.0rc2, I'd say let's just release this in the official 1.9. If you feel strongly that it should be released, I'll let you handle it.

momchil-flex:
Doesn't this just mean re-tagging the pre/1.9.0 branch of the docs repo with the 1.9.0rc2 tag and then maybe manually forcing a build on readthedocs? Or it does, but you're not keen on changing released docs?

tylerflex:
It does but I'm just lazy and not sure if there's a rush.

tomflexcompute:
It looks fine to me to rebuild for 1.9 later. I have another case study using CustomMedium that is ready for Weiliang to review. We can include that as well as Emerson's edge coupler together in official 1.9.

PR #125, opened 2023-02-20
relative links

tylerflex:
Solves the issue of needing to explicitly provide the tag / version in the links in our docs.

I tested this with a local build and it looks good. Eventually (before 2.0) we should go through and test all the links to make sure they all work.


tomflexcompute:
Seems like the links in the notebooks don't work for me. I got the 404 page when I click any links. It works after building the docs?

tylerflex:
Yea, try building them and see if they work for you. Note that you'll get a bunch of warnings about missing links but it still worked for me..

tomflexcompute (approved):
Works for me too!

PR #127, opened 2023-02-21
Tom/optical luneburg lens

tomflexcompute:
Hi @weiliangjin2021 could you review this luneburg lens notebook? Thanks. `CustomMedium` is used in the second part.


tylerflex:
I'll take care of this because I think @weiliangjin2021 has a lot on his plate.

@tomflexcompute this is a really cool demo! I just have a few comments:

* Intro paragraph
	* “high-gain antennas based on Luneburg lens can be achieved by” -> “on Luneburg lens” -> “based on a Luneburg lens design” reads a bit better here.
* text after Cell [8] 
	* Lets either put `quotes` or a link to `FieldMonitor`
* Some of the links break for me. Are they relative links?
* Might be interesting as a final cell to get an overlay of the idealized vs metamaterials Field intensity as a function of x plot at a single z.

@weiliangjin2021 feel free to take a look if you want, but I think we can merge this soon after those comments are addressed. Thanks @tomflexcompute

weiliangjin2021:
Thanks @tylerflex. I'll take a quick look by the end of the day.

tomflexcompute:
Thanks @tylerflex for the comments. They have been incorporated in the revised version. Now all links are using relative links.

tylerflex (approved):

tylerflex:
@tomflexcompute is meatless added twice in the examples.rst? might want to double check this

tomflexcompute:
That's correct @tylerflex . If a notebook can be categorized into multiple categories, I added it to each of them once. 

Do you think it would be a problem? We can restrict each notebook to only one category if you think that's beneficial.

tylerflex:
Hm, I dont think it's a big problem, but it seems a bit strange to me. Were we thinking of having some tags for these notebooks? if so, it might make more sense to have one instance of each in the actual examples list, but then have multiple searchable tags per example? I'm not sure. What do you think is best?

tomflexcompute:
> Hm, I dont think it's a big problem, but it seems a bit strange to me. Were we thinking of having some tags for these notebooks? if so, it might make more sense to have one instance of each in the actual examples list, but then have multiple searchable tags per example? I'm not sure. What do you think is best?

I'm not too sure either. Maybe for now we can stick to one category per notebook. Based on how things are going in the example gallery and knowledge base (websites currently under construction by MC), we can always adjust here.

tylerflex:
ok

weiliangjin2021:
Very nice! Good to merge.

PR #130, opened 2023-02-23
emerson/edge_coupler: reviewed version of Edge Coupler notebook.

e-g-melo:
I've addressed all the comments on the previous PR.


tomflexcompute (approved):
Look great to me. Thanks @e-g-melo for the revision!

tylerflex:
Thanks @e-g-melo , this looks really good.
I just have one minor comment:
there are a few print statements like
```py
print("Estimated FlexUnit cost for " + sim_name + " =", cost)
```
that end up printing out a lot of output, probably a lot more than people need to see and could make it hard to read. 
```
Estimated FlexUnit cost for linear = 0.11789150620527468
```
So I'd suggest using f-string again here and passing a formatting option to only print to 2 decimal places.
```py
print(f"Estimated FlexUnit cost for {sim_name} = {cost:.2f}")
```
should print like 
```
Estimated FlexUnit cost for linear = 0.11
```
Could you make this change and re-run and then I think we can merge this in.
Thanks again for the great work!

tylerflex (changes requested):
see comment just left. Thanks!

tylerflex (approved):

PR #132, opened 2023-02-23
added multiple frequencies to the smatrix

tylerflex:
Two major changes:
1. Smatrix tool returns a DataArray with dims `("port_in", "port_out", "mode_index_in", "mode_index_out", "f")`
2. Smatrix tool accepts `freqs` instead of single `freq`


tylerflex:
Notebook committed directly into develop

PR #134, opened 2023-02-24
Updated current notebooks

tomflexcompute:
1. As Xin requested, the names of competitor software are removed from the benchmark notebooks (HighQ_Ge.ipynb and HighQ_Si.ipynb).

2. Symmetry is added in MMI_1x4.ipynb to speed up the parameter sweep and cut down FlexUnit cost. Some other settings are fined tuned to improve the flow better. Hopefully this saves Momchil's time when rerunning notebooks before release.

3. `;` are added to the rest of the notebooks to avoid console output when plotting. Only the ones that I found, Not sure if all notebooks are fixed but we can do another round of checking in the future.


tylerflex (approved):
Thanks @tomflexcompute !

PR #137, opened 2023-03-01
Simple STL tutorial notebook

shashwat-sh:
The cone-sphere images below cell 10 (the one that shows the wireframe too) for some reason weren't updated correctly in my notebook after I replaced the files in the folder with those of a finer mesh. So the actual images will look a bit different.


tylerflex (approved):
This is a great tutorial notebook, thanks @shashwat-sh!
I think it will be really useful and especially like how you show how to use the origin and other features in practical examples. Just a few minor comments

* Remove reference to binder notebook, we stopped doing those recently.
* Put the quotes in `Pip install “tidy3d[surfacemesh]”` (or is that just on Mac?)
* Getting started: “ww’ll” -> “we’ll”
* Dont forget to put the notebook in `examples.rst`

caseyflex (commented):
This is excellent, thanks Shashwat! I think you did a great job showing how to use STL and also communicating the warning about nonunionized geometry.

A few minor comments:

- I think you need to include the data files in this PR
- You say "Note that if the same STL file contains multiple disjoint objects, it is assumed that those objects will all have the same material properties, because they are treated as a single Tidy3D [Geometry]". I guess you could use the solid_index parameter to create separate geometries for each object, and then assign separate material properties as well.
- In your "Multiple STL solids and multiple files" section, you may want to be consistent in how you order the three examples, so it is more clear
- Could you add a PermittivityMonitor, and suggest that the user use this to verify that the permittivity profile looks good?

tomflexcompute (approved):
Great tutorial for STL import! It demonstrated different common scenarios likely faced by users. Only a few quick comments: 

1. As Casey pointed out, the stl files need to be uploaded as well. I think the best place to add them would be the 'misc' folder, where we store whatever files that are imported in the example notebooks. 

2. Above cell [3] you mentioned defining a point dipole source in the markdown, but a plane wave is added actually. Is that a typo? I don't see any time domain monitors either, w which is mentioned in the markdown too.

tylerflex:
@tomflexcompute I think we've been using the `data` folder for these miscellaneous files. At some point it seems the `misc` folder was added, which does the same thing, but maybe we could combine them into one?

tomflexcompute:
> @tomflexcompute I think we've been using the `data` folder for these miscellaneous files. At some point it seems the `misc` folder was added, which does the same thing, but maybe we could combine them into one?

Indeed that's the case. I think Momchil added the misc folder since the data folder is not tracked by git. I think it's a good idea to have this misc folder for everything that is imported so they don't get mixed with the data downloaded after simulation runs. And yes I agree that we should move every imported file from data to misc.

tylerflex:
ok sounds good, made an issue https://github.com/flexcompute/tidy3d/issues/738 to track this, but maybe @shashwat-sh can you just move all the `data` into `misc` as part of this PR?

tomflexcompute:
> Thanks for the reviews all, I've pushed the revised version and added the data files to `misc`.

Thanks. Do we want to change the import file directories in the associated notebooks?

tylerflex:
I think @shashwat-sh only changed the files from this notebook. but at a later time, we should go through `data` and move the files over and fix their links

shashwat-sh:
Yeah I didn't move the other data because I think we'd need to go through the notebooks carefully and change the import locations in several places.

caseyflex:
`sim_data.plot_field(field_monitor_name="xz_eps", field_name="eps_xx", y=0, val="real", ax=ax2)`
one of these should be sim_data_ref

shashwat-sh:
> `sim_data.plot_field(field_monitor_name="xz_eps", field_name="eps_xx", y=0, val="real", ax=ax2)` one of these should be sim_data_ref

Good catch, thanks, I've pushed an update.

caseyflex:
should we change it to pip install "tidy3d[surfacemesh]=1.10.0rc1" here as well? @tylerflex

PR #139, opened 2023-03-03
Added optical hybrid case study using stl import

tomflexcompute:
Hi @shashwat-sh and @caseyflex could you quickly review this stl import case study? Thanks!

Also fixed a redundancy in the examples.rst. metalens was referenced twice in two subsections. I deleted one of them.


shashwat-sh:
Looks great to me! Pretty compelling in terms of showing off the STL capability. Couple minor things:

- When you import files, e.g. `misc\mmi`, do you generally use the Windows style backslash in other notebooks too? I guess different notebooks then use different conventions? Maybe it doesn't matter thanks to python magic, but I think some notebooks use the unix-style forward slash.
- For the very last simulation, it looks like the CMRR and Imbalance match for I and Q differs more than for the LO output case. Is this expected, or do we expect I and Q to be identical again? Maybe warrants a brief explanation (couple sentences) in either case?

shashwat-sh (approved):

shashwat-sh:
A general comment not related to STL stuff: @tomflexcompute  your point about the metalens notebook is a good one. However, I now realize that one could classify that notebook in more than one way based on how the categories are set up. I wonder if in the future we should somehow add "tags" to notebooks, where the same notebook can have multiple tags. @tylerflex maybe you've already thought about this. Could be a nice addition to Daniil's spice-index idea :)

tylerflex:
I think tags would be great, just not really sure how implement them in the sphinx theme and make them searchable in an easy way.  I'll make an issue to track this feature though

caseyflex (approved):
Thanks for making this very exciting application notebook. Looks great to me!

tomflexcompute:
Thanks @caseyflex @tylerflex  @shashwat-sh . It's ready to merge along with the [stl import tutorial](https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/137).

tomflexcompute:
> A general comment not related to STL stuff: @tomflexcompute your point about the metalens notebook is a good one. However, I now realize that one could classify that notebook in more than one way based on how the categories are set up. I wonder if in the future we should somehow add "tags" to notebooks, where the same notebook can have multiple tags. @tylerflex maybe you've already thought about this. Could be a nice addition to Daniil's spice-index idea :)

- I think both `\` and `/` work fine but for the sake of consistency, we can stick to `/`. 

- For the last simulation, since the signal port is the left waveguide of the bottom MMI, the symmetry is broken so CMRRs could differ. Ideally they are close, which seems to be the case in this design. I've added a few comments about this. 

- Originally I added notebooks in multiple subsections if applicable. However, it seems to cause duplication of entries on the left menu. So I restricted each notebook in only one subsection, but this metalens notebook was missed for some reason. We are building up a knowledge base website with MC and will have an application gallery there to showcase all the examples. Once that is up and running, we can think of better ways to categorize the examples.

tomflexcompute:
The notebook has been updated to adapt to Weiliang's new stl import implementation, which is much faster and more accurate. Should be ready to merge @tylerflex @weiliangjin2021 ? Or do we want to wait until after the solver is built?

tylerflex:
@tomflexcompute . I'll let @weiliangjin2021 comment on that. From my perspective, good to merge as long as the backend can support it. Thanks everyone!

weiliangjin2021:
Green light from me as well.

PR #146, opened 2023-03-15
Updating FAQ section

momchil-flex:
- Added new section on normalization
- Updated sections on units, divergence, and dispersive materials
- Removed section on PML as it was quite stale, and we warn about putting structures exactly at boundaries now

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -29,16 +29,16 @@ For more questions or to purchase Flex units, please contact us at ``support@fle
 What are the units used in the simulation?
 ------------------------------------------
 
-We assume the following physical units:
+We generally assume the following physical units in component definitions:
 
-  - Length: micron (μm, :math:`10^{-6}` meters)
-  - Time: Second (s)
-  - Frequency: Hertz (Hz)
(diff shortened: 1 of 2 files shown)


momchil-flex:
I want to discuss tomorrow how to normalize PointDipole.

tylerflex (commented):
Looks good. Just a note that we have a couple front end issues from a while ago about adding discussions to the docs / faq: 
1. [discussion of auto shutoff](https://github.com/flexcompute/tidy3d/issues/199)
2. [information about PML parameters](https://github.com/flexcompute/tidy3d/issues/135)
I'm not sure how important these are to include, but maybe this is a good time to either add a bit of info on each, or just decide to close those issues.

tylerflex:
For example, it could be nice to document exactly how the `field_decay` value is computed over time.

momchil-flex:
I added two more sections on early shutoff and on not long enough run time. I think the PML parameters is a bit involved/advanced and maybe not needed at this point (probably fine to close the issue too).

momchil-flex:
@xin-flex would you like to take a look?

PR #147, opened 2023-03-16
Added TFSF notebooks

shashwat-sh:
- Added TFSF tutorial notebook
- Updated sphere RCS and plasmonic nanoparticle notebooks to use TFSF and no longer require artificial normalization for comparisons to Mie series


momchil-flex:
Generally great. A few points:

- Remove the `output` directory that comes from local runs
- The dielectric sphere notebook has warnings that the grid is nonuniform in the tfsf region. Let's generally try to avoid warnings unless there's a good reason to keep them in (and in that case, explain to the user why we did that). So let's add an override box there too.
- Not necessarily an actionable item, but I'm wondering why mesh_step_per_wvl of 100 is needed outside of the plasmonic particle. This sounds like way too much. What if you tried decreasing e.g. the auto grid max_scale to 1.2? It would be good to understand what's happening.

shashwat-sh:
Thanks, I've made the changes.

> I'm wondering why mesh_step_per_wvl of 100 is needed outside of the plasmonic particle. This sounds like way too much. What if you tried decreasing e.g. the auto grid max_scale to 1.2? It would be good to understand what's happening.

Yeah the improvement from 50 cells to 100 cells is very tiny. Using a `max_scale` of 1.2 actually helps too - I can get pretty much the same curves with 50 cells per wvl outside the particle and a `max_scale` of 1.2. Making the mesh coarser than that does start to increase the deviations a little bit.

PR #154, opened 2023-03-24
Added graphene and microwave FFS case studies

tomflexcompute:
Adapted both 2d material case studies to use auto grid. 

Override structure is demonstrated in the frequency selective surface notebook.


momchil-flex (approved):
These are great!

PR #157, opened 2023-03-27
Moving imported files from data folder to misc folder

tomflexcompute:
Moved all imported files from the data folder to the misc folder. Note that the .json files are not moved as I think they are better in the data folder. 

Another note is that in the PlasmonicNanoparticle.ipynb, three .txt files are imported. However, I only found the last one in the data folder. It seems like the other two were never uploaded to the repo. As a result, I couldn't run this notebook either. Could you please double-check? Maybe you have these files in your local data folder?

image


momchil-flex:
Yeah, no worries, I had already asked Shashwat for these files. Since I'm in the middle of re-running notebooks already, I'll just incorporate these changes in my upcoming PR, but this is very useful for showing where exactly the changes need to happen.

momchil-flex:
Changes incorporated in https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/158

PR #162, opened 2023-03-29
Update AutoGrid and CustomMedium notebooks

weiliangjin2021:
- In AutoGrid notebook, update the import of mode solver; in the first plot, adds `ax = ...` to remove output of ax info
- In CustomMedium, rerun to remove all `ax= ...` output.


tylerflex:
Looks good, just a few things:
1. The CustomMedium notebook looks like it still uses `;`. Let's just throw in `plt.show()` instead or assign output to `ax` or `_`.
2. The `AutoGrid` notebook has a warning about no sources in the simulation. Can we throw in a line in the markdown cell after that saying `> Note: Tidy3D is warning us that our Simulation does not contain sources. In this case, since we are using the simulation as a demonstration of the meshing process and are not running any simulations, we may safely ignore this error.
3. Might want to add a similar message about the last warning in the Autogrid notebook. In the metalens, I added this message `The warnings are just letting us know that we are using uniform grid along x and y (per our spec) even though the override structures have some extent in these dimensions. We can ignore as this is intended.`

tylerflex (approved):
see my comments but after this, it seems ready to merge.

PR #163, opened 2023-03-29
tyler-run notebooks for 2.0

tylerflex:


momchil-flex:
Tried to look through, even though it's hard even with github's new notebook diff functinality. Just a few comments:
- In ParameterScan `config.logging_level = "error"` is commented out: should it be completely removed if not needed?
- In AdjointPlugin notebook 3:
  - Cell 18 needs a print statement
  - In cell 7, when you make a fig size too big, it still gets shrunk to your page width, but the font size gets smaller. It's usually best to keep the figsize width up to like 14 or so. The figure would probably be better if the fig size was something like (14, 8). This comment may apply to other figures / notebooks too...
 
And a question: how did you verify links? Seems like you fixed some broken ones but I'm not sure if there's an easy way to check them...

tylerflex:
Thanks, I'll fix those two notebooks.

To verify links, it's a bit rudimentary, but I just checked that the existing links worked on the published docs (not in the jupyter notebooks). I actually found a couple broken ones this way.

tylerflex:
Ok I made the changes, I guess feel free to merge if you think it's good to go.

momchil-flex:
Sorry, regarding the plot, it seems like `fig, axs = plt.subplots(num_modes, 3, figsize=(14, 10), tight_layout=True)` looks good with no overlapping of text. At when I try on my machine.

tylerflex:
Ok I set it to that in the new version and ran up to the plot

momchil-flex:
Looks good, did you want to squash your commits before merging?

tylerflex:
squashed

PR #164, opened 2023-03-29
weiliang-run notebooks for v2.0

weiliangjin2021:


tylerflex (approved):
Looks good, only comment: In the fitter [15], can we address the warning about the points being outside of the frequency range? maybe we need to modify the fitter parameters?

After this you can merge, thanks!

tylerflex (changes requested):
see comment above

weiliangjin2021:
I think this is because of the change of speed of light in the recent frontend, while the webservice fitter is still using the v1.8.0. So this leads to a tiny change of frequency range. Let me try to update the webservice and rerun this notebook.

PR #166, opened 2023-03-30
Momchil updating notebooks for 2.0

momchil-flex:


tylerflex (changes requested):
Looks good, just a few comments:
* Will need to fix the submodule conflict I guess.
* TFSF notebook
   * I think we should add a note after the warnings about override structure eg `> Note: The warnings are just letting us know that we are using uniform grid along x and y (per our spec) even though the override structures have some extent in these dimensions. We can ignore as this is intended.`
   * The tile "Make the reference simulation with the plane wave source, and an empty simulation with a plane wave source to allow computing scattered fields." seems excessively long.
   * The bullet points [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/TFSF.html#TFSF-source-in-a-layered-substrate) dont render correctly in the actual html notebook, I think the fix is to add a newline between each bullet.
After this feel free to merge it.

PR #167, opened 2023-03-30
Basic mode source notebooks for 2.0

dbochkov-flexcompute:


tylerflex (changes requested):
Modal_sources_monitors
   * in cell [9] there is a warning about field decay. After this, can we write a quick note mentioning why we decide to ignore this warning in the notebook?
Mode Solver
    * in cell [3], we get a warning about no sources in the simulation. Can we add a note about why this is ok for mode solver simulations? eg `Tidy3D is warning us that the simulation does not contain a source. However, since this simulation is used to construct the mode solver and will not be run directly, we can ignore this warning.`

Similar comments about warnings in the other 2 notebooks.

This is pretty minor but I think will help new users if they see a warning in their own code so they have a sense for when it's ok to ignore them.

After these changes, feel free to merge this PR, thanks!

PR #170, opened 2023-03-30
Updated notebooks for 2.0

tomflexcompute:
- Changed Plugin imports

- Changed plot_field method arguments (`int` -> `abs^2`)

- Added `plt.show()` to replace `;`.

- Changed `FlexUnit` to `FlexCredit`.


tylerflex (changes requested):
Thanks Tom!

* 8 channel
    * maybe after explaining about the logging level being set to ERROR we should set it back to WARNING once the mode solver stuff is through? That way we encourage users to leave it on in case it catches warnings in the main simulation? If it will take too much time to re-run we can ignore this maybe.
* Do we want to be `verbose` in the runs for the following notebooks? (I'd say only leave it False if the runs are sufficiently fast that a user wont think the program is stuck).
    * Dielectric Metasurface?
    * Eueler?
    * Graphene?
    * Microwave?
* THz
    * In the last two `sim_data.plot_field()` I think we should supply the keyword arguments to the `'field', 'E', 'abs^2'` values, just so people can learn what they are.
     * This applies to other notebooks too, (Y junction, etc) if you think it's not worth re-running, I think we can leave this to later. It might exist in other notebooks too. We can make an issue on the front end if so.
 * Y junction
     * cell [10] terminates with `;`. Let's remove and throw in `plt.show()` instead. Check other cells.

PR #172, opened 2023-04-05
Add group index calculation example to ModeSolver

lucas-flexcompute:
Goes together with https://github.com/flexcompute/tidy3d/pull/828


tylerflex (approved):
Looks good to me, we can merge this into a new 2.1 docs branch once the front end is merged. Thanks!

PR #178, opened 2023-04-14
Add waveguide plugin example notebook

lucas-flexcompute:
Example notebook to showcase the uses of the waveguide plugin.


tomflexcompute (commented):
Thanks @lucas-flexcompute for making such a nice tool. A few minor comments below for your consideration.

1.	Usually we only capitalize the first letter of the first word on the notebook title so to follow the tradition, we can make it “Waveguide plugin”. Alternatively we can make it slightly more informative as something like “Waveguide plugin tutorial” or “Waveguide plugin demonstrations”.

2.	Why is the field value slightly asymmetric in the slot waveguide shown in Cell[21]?

3.	In the last plot, the effective index doesn’t seem to converge even with grid=resolution=45. I think this is a common issue in ModeSolver that has been brought up many times by our users. Maybe we should add a comment here and explain why Python side ModeSolver is not that accurate and if you need accurate results, you should use ModeSolverMonitor etc.

lucas-flexcompute:
> Usually we only capitalize the first letter of the first word on the notebook title so to follow the tradition, we can make it “Waveguide plugin”. Alternatively we can make it slightly more informative as something like “Waveguide plugin tutorial” or “Waveguide plugin demonstrations”.

Sounds good! I've renamed it to "Waveguide plugin demonstration".

> Why is the field value slightly asymmetric in the slot waveguide shown in Cell[21]?

From what I can see, the mode returned from the mode solver is asymmetric. It doesn't seem to be an artifact of the mesh or the plot. Note that the field plot in the center is also asymmetric. The asymmetry remains even for higher grid resolutions.

> In the last plot, the effective index doesn’t seem to converge even with grid=resolution=45. I think this is a common issue in ModeSolver that has been brought up many times by our users. Maybe we should add a comment here and explain why Python side ModeSolver is not that accurate and if you need accurate results, you should use ModeSolverMonitor etc.

True, I've added a comment along these lines. Let me know if you have any suggestions there.

tomflexcompute (approved):
Looks great. Thanks for the revision. 

Before you merge it, could you add an entry for it in the examples.rst file? I think it can go under the Advanced Structure Definition category or another that you can fit better. Thanks!

tylerflex:
@lucas-flexcompute / @tomflexcompute we may need to add the waveguide plugin to api.rst as well. see the other examples.

https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/develop/docs/source/api.rst#plugins

Can @lucas-flexcompute make another PR for that? thanks

PR #179, opened 2023-04-14
Added Fresnel lens case study

tomflexcompute:
Customers have kept asking to see simulations with large domains to show our true capability. Someone happened to ask about Fresnel lens example. This Fresnel lens case study is meant to demonstrate how large of a model we can run and how fast we can run it. The simulation domain is over 100 wavelengths in all three directions and only takes about 6 mins.


momchil-flex:
This seems similar to this example: https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/ZonePlateFieldProjection.html

Just bigger, and a better lens probably. However, just as in the zone plate example, it probably doesn't make sense to simulate the entire space past the focal point in z. Practically, the user should simulate a thin slice, and use field projections. Of course we could leave it as is just to demonstrate our capabilities, but it's a bit off-putting that technically you could do an equiivalent cheaper/smaller simulation. What do you think?

tomflexcompute:
> This seems similar to this example: https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/ZonePlateFieldProjection.html
> 
> Just bigger, and a better lens probably. However, just as in the zone plate example, it probably doesn't make sense to simulate the entire space past the focal point in z. Practically, the user should simulate a thin slice, and use field projections. Of course we could leave it as is just to demonstrate our capabilities, but it's a bit off-putting that technically you could do an equiivalent cheaper/smaller simulation. What do you think?

Thanks for reviewing. The issue with using field projection right away is that initially we don't know exactly where the focus is. One can set up a FieldMonitor close to the lens and do field projections locally after simulation to try different projection distances to determine the focus. However, local field projection with such a large cross-section is very slow. I tried to do it and each projection took something like 50 mins even with a somewhat reduced resolution. In comparison, the full domain simulation only took 6 mins. Although it costs a bit more, some users would rather spend a bit more to save time. Of course one can also set up multiple FieldProjectionMonitors set to different projection distances and do server-side projections, but that still takes quite a bit of trial and error to nail down the exact focus position. In addition, some people might be interested in how the beam converges so they want to look at the field distribution from the side (xz or yz view), which is hard to obtain from field projection.

We can also demonstrate how to do server-side projections as a comparison after we know where the focus is from the full domain simulation. I didn't include it since that has been demonstrated in both the zone plate and metalens notebooks.

momchil-flex:
I see, sorry I didn't read the notebook in more detail at first, you kinda explain the same thing there.

I am still wondering about this though. I would think it's possible to recreate the xz-plane results using projection, rather than having to place xy planes a certain distance away. @shashwat-sh is it possible to basically reproduce the xz fields shown in the notebook using a thing simulation in z and a FieldProjectionCartesianMonitor in the xz plane? The one thing that's confusing to me is if the near field monitor is 2D and in the xy plane, can the `proj_axis` still be set to 1?

tomflexcompute:
> I see, sorry I didn't read the notebook in more detail at first, you kinda explain the same thing there.
> 
> I am still wondering about this though. I would think it's possible to recreate the xz-plane results using projection, rather than having to place xy planes a certain distance away. @shashwat-sh is it possible to basically reproduce the xz fields shown in the notebook using a thing simulation in z and a FieldProjectionCartesianMonitor in the xz plane? The one thing that's confusing to me is if the near field monitor is 2D and in the xy plane, can the `proj_axis` still be set to 1?

Yeah I think the xz-plane field profile can be recreated purely from field projection. Essentially that requires doing field projection multiple times at different projection distances (in the order of 100 projections is needed in this case to construct the same result). This will inevitably take a long time on a local machine. On my computer for example, it would take well over 15 mins. 

One thing we can potentially change to make it a viable option would be to allow `proj_distance` in `FieldProjectionCartesianMonitor` to be a list. This way only one `FieldProjectionCartesianMonitor` needs to be defined and projections at multiple distances can be done on the server easily. @shashwat-sh would this be possible?

momchil-flex:
What I'm proposing is rather proj_axis=1, which makes the monitor defined x and y being the global x and z.

tomflexcompute:
Not sure if it's supported? How is the projection distance defined in this way?

shashwat-sh:
>  @shashwat-sh is it possible to basically reproduce the xz fields shown in the notebook using a thing simulation in z and a FieldProjectionCartesianMonitor in the xz plane? The one thing that's confusing to me is if the near field monitor is 2D and in the xy plane, can the proj_axis still be set to 1?

It might be possible, but the way things are set up, it's not clear how `projection_distance` would be defined in this case, since it is assumed to be a constant with respect to the near-field monitor's local origin. Might be worth a try @tomflexcompute by setting `projection_distance` as the distance between the near-field monitor's center and the distance to far-field xz plane along the `y` direction, and setting `proj_axis` to 1 as Momchil suggested.

> This will inevitably take a long time on a local machine. On my computer for example, it would take well over 15 mins.

To use the approach of defining multiple far-field planes (note: this is not what Momchil was suggesting), I think it should be quite feasible computationally using server side projections?

> One thing we can potentially change to make it a viable option would be to allow proj_distance in FieldProjectionCartesianMonitor to be a list. This way only one FieldProjectionCartesianMonitor needs to be defined and projections at multiple distances can be done on the server easily. @shashwat-sh would this be possible?

Yeah it's possible, and we did try this at one point, but at the time it led to a fair bit of messiness in the data handling when converting between coordinate systems, and didn't seem worth it because there's no computational advantage.

> This way only one FieldProjectionCartesianMonitor needs to be defined and projections at multiple distances can be done on the server easily.

Just fyi, computationally, one monitor with multiple `projection_distance`s vs. multiple monitors with one `projection_distance` won't really make a difference in time taken, because the server does a straight-up loop through observation points. Of course, managing the data afterwards would be much easier if it was all packaged together.

shashwat-sh:
And, with Momchil's idea, the `x` and `y` observation points in the far field monitor are actually the global `x` and `z` coordinates. Also, regardless of the approach I think you'd have to set `far_field_approx` to `False` and use the exact projections, if the smallest projection distance is comparable to the structure size.

shashwat-sh:
By the way, if Momchil's idea doesn't work, I think maybe it's worth the effort for me to add this ability in the near future? Seems quite useful.

tomflexcompute:
Thanks for the clarification. Will give it a try and update you guys.

momchil-flex:
> It might be possible, but the way things are set up, it's not clear how `projection_distance` would be defined in this case, since it is assumed to be a constant with respect to the near-field monitor's local origin. Might be worth a try @tomflexcompute by setting `projection_distance` as the distance between the near-field monitor's center and the distance to far-field xz plane along the `y` direction, and setting `proj_axis` to 1 as Momchil suggested.

Yeah that's exactly what I had in mind, so let's say `proj_distance = 0`.

Generally, I see no reason why something like this shouldn't work. I would expect that with the far field approximation turned off, I should be able to get fields anywhere I want (they just may be inaccurate on very short distances). If that's not the case, then we should probably reconsider the projection monitor definitions yeah.. But again, I don't see why this wouldn't work, unless there's something specific that's e.g. validated for a 2D monitor.

shashwat-sh:
Agreed, there's no mathematical or physical reason it won't work, it's just
a matter of whether such a setup is handled correctly in the implementation
right now. If not, it shouldn't be hard to modify and make this work
without needing any API changes.

tomflexcompute:
@momchil-flex @shashwat-sh Setting `proj_axis` to 0 or 1 does work. Thanks for the suggestion. I've added another section to demonstrate this. However, the field from projection seems to be 100 times smaller than that in full simulation. Could you see why that's the case?

shashwat-sh:
Nice! Could you check a couple things:
- If you sample the projected "far" fields on an xy plane, say near the focus, but using the "conventional" setup where the `proj_axis` is `2`, then do the field magnitudes match? There are several other tests where projected field magnitudes match the ones computed directly from `FieldMonitor`, and the only difference here seems the odd setup, so I'm wondering if it's specific to that.
- It looks like you're plotting the intensity - do the raw fields themselves also differ a lot in magnitude?

shashwat-sh:
Actually, I'm also wondering if the `ys_far` should be `ys_far = np.linspace(-110,0,200)` since it corresponds to global `z` values in the bottom half space?

momchil-flex:
> Actually, I'm also wondering if the `ys_far` should be `ys_far = np.linspace(-110,0,200)` since it corresponds to global `z` values in the bottom half space?

Oh, funny, so the projection results may be the reflected light which is also focused!

tomflexcompute:
Thanks for spotting this! Indeed I was looking at the reflected focus :sweat_smile: and since the reflection is about 1%, the field is off by a factor of 100 and intensity by 10000. The updated notebook has this corrected and it works beautifully. 

A minor issue is that the projected field at y=0 seems to be cut off. Is this related to the symmetry used? I remember another situation a while back where using symmetry caused some field value issues at the symmetry plane.

shashwat-sh:
Hmm, is it possible it's just because the `xs_far = np.linspace(-50,50,200)` is not sampling at 0, so you're seeing the samples at -0.25 and 0.25, but not at 0? Maybe you could try more `xs_far` points, or have an odd number of sample points / plot as dots instead of lines?

I'm not sure if symmetry is the issue, but I guess if the simulation doesn't take too long you could try without and see if it improves (if what I said above isn't the problem)

shashwat-sh:
This is an awesome validation of field projections btw, I have to admit I'm almost surprised it matches this closely...

tomflexcompute:
Thanks fo the suggestion. Indeed that has something to do with the field projection grid. I've used a finer grid in the updated version. The field intensity at the center is still a tiny bit off but that's likely due to interpolation so I won't be too worried about it.

shashwat-sh (approved):
Look great to me! A few very minor things:

- Small typo in the figure's legend below cell 4: `discritized`
- Right above cell 11, maybe "setting" shouldn't be in single quotes, where it says `setting proj_axis=0`
- In cell 11, the comment `proj_axis=0,  # axis along which to project, in this case z` should probably be updated to say `x` instead of `z`

momchil-flex (approved):

momchil-flex:
I think develop is fine, thanks.

PR #186, opened 2023-05-01
Added SWG polarizer case study

tomflexcompute:
Could you guys review this SWG polarizer case study? Thanks in advance.


lucas-flexcompute (approved):
Looks great! I have only a couple of points regarding the introduction:
1. I wouldn't use the word "novel". The example (and design) will get older and we probably don't want to revisit the examples to change that.
2. I always prefer the DOI links (https://doi.org/10.1364/PRJ.7.001432 in this case), because they are future-proof. In fact, I would format the reference as `Hongnan Xu, Daoxin Dai, and Yaocheng Shi, "Anisotropic metamaterial-assisted all-silicon polarizer with 415-nm bandwidth," *Photon. Res.* 7, 1432-1439 (2019), [DOI: 10.1364/PRJ.7.001432](https://doi.org/10.1364/PRJ.7.001432)`, that is with the journal title emphasized and the link only over the DOI.
3. It's probably a good idea to describe the acronym SWG on the first use. I've see it being used in other context with different meanings ("strip waveguide", "silicon waveguide", etc)

e-g-melo (approved):

PR #188, opened 2023-05-08
Added tmm demo case using a directional coupler example

tomflexcompute:
This case study breaks down into two parts. The first part is using transfer matrix to analyze the directional coupler design and get an estimate of the transmission. However, I found the equations in the original paper problematic. In the notebook, I used equations that I think are correct and the result also makes sense. More specifically, the paper uses
image
but I don't see why we should use the inverse matrix of `$P_t$` instead of just `$P_t$`. Please double check if you have time and let me know what you think. 

The second part is just regular FDTD of the optimized design so there is nothing particularly noteworthy. 

Thank you in advance!


lucas-flexcompute (approved):
Very interesting example! I liked it a lot. You are right, the second matrix in the equation should no be inverted. I have only a few comments:

- Why is `fwidth` defined with 40% of the frequency window and not the full range?
- I'd remove "and only cost a small amount of FlexCredits" from the paragraph before the taper simulation.
- In cells 12 and 14, I'd add a line plot for θ_t1 and θ_t2, just for illustration.
- Typo: "we are ready to assemble the *matrices* and…"
- You mention using a grid search to optimize both lengths. Would it be possible with adjoint? Or even using an external gradient descent? Either would be more efficient than a grid search. Maybe you could mention that the optimization can be performed any way the user prefers, or simply using a grid search.

e-g-melo (approved):
Hi Tom! Congrats, the notebook is very interesting.

Some minor comments:
1) The matrix P and Pt have "-0" values.
2) It would be interesting to define the splitting ratio bandwidth in terms of -3dB +- x%. In this way, you could include upper and lower lines at -3dB+x and -3dB-x values indicating the bandwidth limits in the splitting ratio figures. Then adjust the y-limits to zoom in on the splitting ratio curves.

Best,
Emerson

tomflexcompute:
Thanks @lucas-flexcompute and @e-g-melo for the very constructive comments! 

For the `fwidth`, I think our definition is such that 0.5*(max(freqs) - min(freqs)) will cover the whole frequency range with good signal-to-noise. Usually 0.3 or 0.4 is sufficient too. I'm using a slightly narrower `fwidth` mainly to avoid warning about frequency outside of the library silicon material frequency range.

tylerflex (approved):
This is a very good notebook @tomflexcompute. Does an amazing job introducing the waveguide plugin and how to use it to great effect!

I just had a few cosmetic changes but other than that, looks good to me. Feel free to merge this into `develop` when the changes are made and it should automatically propagate to the docs about 15 minutes after if the changes I made yesterday worked as expected. Please verify.

Thanks!

-> “As a passive device, DC allows” -> “As a passive device, a DC allows”
-> “elegant, yet simple, structure” -> “elegant, yet simple structure”
-> Some of the phase plots are slightly confusing due to discontinuities. Maybe we can apply [unwrap](https://numpy.org/doc/stable/reference/generated/numpy.unwrap.html) or similar if possible?

tomflexcompute:
Thanks again @tylerflex @lucas-flexcompute @e-g-melo for spending the time reviewing it. The changes have been made and it will be merged now. Will check the docs site to see if it's automatically updated.

PR #190, opened 2023-05-12
Added animation tutorial

tomflexcompute:
Making animations out of FDTD is quite common and it turns out many users don't know how to do it, so I made this simple demo. The animation is displayed within the notebook using `HTML`, which I suppose will buit fine on our documentation website?


tomflexcompute:
Actually the animation doesn't render in Github. Not sure if that's an issue?

tylerflex:
I'm not sure it will render in the readthedocs hosted site either but we could try to build it and see. I'll take a look and run the notebook in a bit.

tomflexcompute:
Thanks @tylerflex . I just tried to build it locally and it actually looks fine. Feel free to double check.
image

tylerflex (approved):
This is awesome! I always wanted this but never figured out a great way. This notebook should be really useful to people. Just a couple things:

"display the animation directly in Jupyter Notebook" -> "display the animation directly in a Jupyter Notebook"

in the sim_data plotting, does `.plot_field()` work in time domain? if not, you might want to consider also passing `x='x', y='y'` kwargs into the DataArray plotting function to make it not rotated 90 degrees. See cell [26] of [this](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/VizData.html). Also apply this to the animation maybe.

Thanks tom! feel free to merge this into develop when these are addressed.

PR #193, opened 2023-05-16
Fully anisotropic basic demonstrations

dbochkov-flexcompute:
Two notebooks: one demonstrating the general usage of fully anisotropic materials and another one on gyrotropic media

Todo's:
- [x] remove `solver_version`
- [ ] check links are working once docs are rebuilt


tomflexcompute (changes requested):
Thanks @dbochkov-flexcompute for the great work! Both notebooks are great demonstrations of the fully anisotropic medium. A few comments below for your consideration:

FullyAnisotropic.ipynb
1.	Usually for notebook titles we only capitalize the first letter of the first word, i.e. “Fully anisotropic materials”. Same for the other notebook.
2.	“Tidy3D capabilities” ->“Tidy3D’s capabilities”.
3.	“dieletric” -> “dielectric”.
4.	“transofrmation” -> “transformation”.
5.	“differenes" -> “differences”.
6.	Currently we use a general link when linking to documentation. This is to avoid using the fixed docs version. For example, the link https://docs.simulation.cloud/projects/tidy3d/en/latest/_autosummary/tidy3d.FullyAnisotropicMedium.html
can be replaced with
../_autosummary/tidy3d.FullyAnisotropicMedium.html 
This way, the link will be associated with different versions (latest, stable, 1.9.0, etc) when it’s open in different versions of the docs website. Same for the other notebook.

Gyrotropic.ipynb
1.	Just for my own curiosity, I wonder if td.UniformCurrentSource and td.PlaneWave are the same? Is there a reason to choose one over the other?
2.	“symetric” -> “symmetric”.
3.	“beggining” -> “beginning”.
4.	“discertization” -> “discretization”.
5.	“comparision” -> “comparison”.
6.	“definining” -> “defining”.

In addition, before it's ready to merge, could you update the examples.rst and api.rst as well? In examples.rst, maybe both notebooks can go to Advanced Structure Definition? Otherwise feel free to create new categories if you see fit.

dbochkov-flexcompute:
Thanks for reviewing, Tom! Just fixed the things you've mentioned. 

Regarding:
> Just for my own curiosity, I wonder if td.UniformCurrentSource and td.PlaneWave are the same? Is there a reason to choose one over the other?

we now actually prohibit injection of plane waves into anisotropic materials because in general it requires additional calculations to find correct propagating normal modes and to decompose a specified polarization into normal modes.

tomflexcompute (approved):
Thanks! Looks great.

momchil-flex:
Thanks @dbochkov-flexcompute for these and @tomflexcompute for the thorough review. A few more comments from me:

FullyAnisotropic notebook:
- (nit) `Tidy3D's capabilities include modeling of non-dispersive fully anisotropic materials (see [API reference] for FullyAnisotorpicMedium)` - maybe just `Tidy3D's capabilities include modeling of non-dispersive [fully anisotropic materials].` There are also other instances of `API reference` being linked instead of just the component in question. Similarly, `Tidy3D provides a convenience class RotationAroundAxis ([docs])` can just be ``Tidy3D provides a convenience class [RotationAroundAxis]`.
- In cell 13, add figsize, something like `fig, ax = plt.subplots(1, 3, figsize=(10, 3))`

Gyrotropic:
- Similarly, add figsize to your figures to expand them a little vertically
- A number of sentences that precede a code cell are missing a closing period. We usually add those unless the sentence is continuing after the code cell.
- "Note that because of made approximations the simulation results will be the most accurate" -> "Note that because of this approximation, the simulation results will be most accurate".
- Add some more explanation as to why we use uniform current source. You can explain why plane waves are generally prohibited as getting eigenmodes is complicated. For example, the very fact that we are observing gyration of the polarization here means that a plane wave of a given polarization is not an eigenmode.
- Also explain a little why we need a reference simulation (is it just to get correct normalization?) This also probably plays together with the explanation of using UniformCurrentSource: you can say that in a regular material, the difference between that and a normal-axis PlaneWave is that the former is biderictional, and not normalized to unit flux.

dbochkov-flexcompute:
@momchil-flex thanks for the suggestions, applied them all

> Add some more explanation as to why we use uniform current source. You can explain why plane waves are generally prohibited as getting eigenmodes is complicated. For example, the very fact that we are observing gyration of the polarization here means that a plane wave of a given polarization is not an eigenmode.

I decided not go into details of what it takes to insert a planar unidirectional pulse into anisotropic media, since we don't currently support it not because it's fundamentally difficult, but just because it has been implemented. Just mentioned that we are not supporting it yet: 
_Since Tidy3D does not currently support injection of [unidirectional plane waves](../_autosummary/tidy3d.PlaneWave.html) into anisotropic material, we use [a uniform current source](../_autosummary/tidy3d.UniformCurrentSource.html) to initiate a planar (bidirectional) propagating pulse._

> Also explain a little why we need a reference simulation (is it just to get correct normalization?) This also probably plays together with the explanation of using UniformCurrentSource: you can say that in a regular material, the difference between that and a normal-axis PlaneWave is that the former is biderictional, and not normalized to unit flux.

Not sure if we can call it normalization, but what I do: take results of the reference simulation, apply polarization rotation manually as predicted by theory, and compare that to the gyrotropic simulation. I added the following explanation in the beginning: 
_Let us now demonstrate simulation of plane wave propagation inside a gyrotropic material. Specifically, we will consider two simulations, one with $\boldsymbol{g} = 0$ and another one with $\boldsymbol{g} \neq 0$. In the first, reference, simulation the injected plane wave is expected to preserve its polarization, while in the second case the polarization of the injected pulse will undergo rotation due to the gyrotropic properties of the background medium. The rotation degree can be predicted by analytical expressions. In section [Results](#Results) we apply it to the results of the reference simulation and compare with the results of the gyrotropic simulation._

dbochkov-flexcompute:
rebased onto `pre/2.2`

PR #196, opened 2023-05-18
Added bilevel PSR notebook and modified the previous compact PSR note…

tomflexcompute:
- Added bilevel PSR notebook.
- Modified the previous compact PSR notebook slightly to make the bending part smoother.

This example is suggested by one of our users. Originally I was helping him to set up the simulation but it works very well so I figure we can make it a case study for us too (the user also gladly agrees). We plan to make regular social media posts about our case studies that are large and interesting. This half mm device will be featured in the first post. Even though the length is above 0.5 mm, the simulations only take 10 mins so I hope it's ok to keep it in our regular case studies. For even larger models, we might want to keep them in a different place so they don't have to be rerun regularly and maybe only once in a while @momchil-flex ? When there is a major release like 3.0?


momchil-flex:
Awesome! One small comment on the text - are the asterisks around the bold words intended, or are you just using `***bold***` instead of `**bold**` accidentally?
image

Otherwise, I kinda feel like this example should be already separated somehow. It seems like each of the two simulations costs 15FC. Or maybe for now we can at least put a message on top that this example exceeds 10FC? Like up to now we've only had this message for 1FC. When re-running notebooks, we have an easy way to exclude whatever we like, so it's not that much of an issue.

tomflexcompute:
Thanks @momchil-flex . That's a great point! I'll put a comment on top saying "Running this notebook will cost 20+ FlexCredits.", so trial users and unpaid academic users with 10 FC shouldn't try to run it and use all FC instantly. 

And yeah somehow I accidentally put an extra * for no reason.

e-g-melo (approved):
This case study is really nice! Very good to demonstrate how fast Tidy3D is. Congrats Tom!

Just two minor comments:
1) w5 and w6 are overlapping in the figure.
2) It would be interesting to have a mode analysis (if you have time) showing why the TE0 input is transmitted in the top waveguide staying as the TE0, and the TM0 input is rotated to TE0 at the bottom waveguide.

Best,
Emerson

tomflexcompute:
Thanks @e-g-melo @momchil-flex for the review and great suggestions. The changes have been incorporated. 

It turns out both *** and ** work correctly as bold on my notebook locally. Maybe it's a windows thing. 

For the overlapping arrows, since both of them indicate the widths at the same x, I don't have a good way to mark them without overlapping. I used different colors for the arrows to separate them. Hope it works.

tylerflex (approved):
look good, I'll merge into develop

PR #202, opened 2023-05-23
Feature/notebook sync

lius-magic:


tylerflex:
Looks good to me. By the way, I built a version from readthedocs.com and configured it to be public, active, and hidden. That way I could look at the resulting build to make sure it rendered properly before merging into develop branch.  Here is that [build](https://docs.flexcompute.com/projects/tidy3d/en/feature-notebook_sync/). Let's do this for any new features added to the repo because develop branch is live and gets displayed when the user navigates to our public docs.  Also, I noticed earlier a commit was made directly to develop to add the GitHub action, because the develop branch is live, I'd ask that we always make a PR and let me or someone else from the EM team merge it. If it is easier, we can create a staging branch for the docs before they go live. For example `develop`, `latest` branches where `latest` is live but `develop` is just a staging branch. Thanks

PR #205, opened 2023-05-24
Modified a few titles and metadata to test if it works

tomflexcompute:
Before changing all the notebooks, I would like to first modify a few titles and metadata to test if it works.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -1,7 +1,7 @@
 Examples
 ========
 
-Introduction
+Tidy3D Introduction
 ------------
 
 .. toctree::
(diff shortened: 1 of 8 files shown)


tylerflex:
@tomflexcompute do you want this merged into develop? If not, you can log into `readthedocs` site and see the result of the build. You can go to "versions", find this branch, find your branch, and change it to "active", "hidden" and "public". 
image

Triggering a build of that branch at "builds" will then let you view the final version in your browser.

tomflexcompute:
@tylerflex I'm mainly trying to see the build on our own site that Yongwei implemented. I think it's only linked to develop?

PR #208, opened 2023-05-25
add anderson localization notebook

tylerflex:
I'm not sure where to insert metadata but here's what I would put

## Title
Demonstration of Anderson Localization of Light in 3D using Tidy3D FDTD

## Description
This notebook computes light scattering in random media using FDTD code and demonstrates 3D Anderson localization of light.

## Keywords
Optical, Anderson localization, random, medium


tomflexcompute (approved):
Very interesting simulation! A few minor things:

1. Under "Dielectric Sphere Transmission": "transmited" -> "transmitted"; "dilectric" -> "dielectric"; Is "permiability" correct? 
2. The cost of the simulation is probably somewhat high? Might want to add something like "This notebook contains large simulations. Running the entire notebook will cost about 30 FlexCredits." similar to https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/BilevelPSR.html. If we let users run it for free (not sure how it can be implemented?), then maybe it's not necessary. 
3. It looks fine to me to put this under "Optical Scattering and Far-field Radiation", but if you can think of a better category, feel free to create a new one. Our current categories are a bit limited and at some point we will have to expand it. For some examples that are kind of unique, I'm thinking about creating a "Others" category.

tomflexcompute:
@tylerflex the metadata can be added in Jupyter Lab. The regular Jupyter Notebook doesn't allow this, at least I didn't find a way to do it either. Here is a screenshot of what it looks like
image

If you prefer, I can add the metadata for you too.

tylerflex:
Hm, all I see is this, wonder why you have so much metadata in yours? maybe could you add it if you dont mind?
image

Regarding the category, it's a bit hard to figure out what to put this in. Maybe let's go with optical scattering for now and then once we have some more "other" examples, we can create a new category?

Thanks for the review! just committed the changes.

PR #213, opened 2023-05-31
Inclusion of grating coupler adjoint optimization.

e-g-melo:


tylerflex:
This is great! I already gave most of my comments directly, so I'm approving pending only one thing:

* When I suggested the warning about credit usage, I just made up the "30" FlexCredit number. Could you check to see how many credits this notebook would require and replace this number?

Thanks Emerson!

tylerflex (approved):

tomflexcompute (approved):
Thanks @e-g-melo for the great case study. This is a fantastic demonstration! The result of the final design is excellent too. I think it’s ready to go with a few typo fixing. Just for my curiosity:

1.	How long exactly does the optimization take? It says several hours, but is it close to one hour or ten hours? 

2.	It is mentioned that adding a mirror can reduce back reflection. Could you explain a bit more about this? Is it something people implement experimentally?

Some typos that I found:
Grating coupler inverse design configuration:
“desing” -> “design”
“bellow” -> “below”
“Grid size in within the design region” -> “Grid size within the design region”

Inverse design optimization set up:
“outpput” -> “output”

Fabricability improvements:
“hiperbolic” -> “hyperbolic”

Optimization:
“loggin” -> “logging”

Optimization results:
“coulpler” -> “coupler”

Some additional minor comments for your consideration. Rerunning the notebook can take a long time so if rerunning is required, please ignore them. 

1.	1000 is used to represent infinity when using Box.from_bounds. I usually define an “inf_eff=1000” variable and use it repeatedly in case the value needs to be changed. Not really necessary I guess but it looks a bit cleaner.   

2.	The cell number starts from [29], which can be a bit strange to readers. Not a huge issue. 

3.	For the H1 header, only the first word is capitalized, which is exactly what you’ve done here. For the H2 headers, usually we capitalize all words (using the typical title format). I don’t think it needs to be strictly enforced but having consistent formatting throughout all notebooks would be nice. Also usually we don’t put “.” in the end of the headers. 

4.	The link for JaxCustomMedium is the full link to the latest. Could you change it to ../_autosummary/…?

tomflexcompute:
Forgot to mention that going forward we will be adding metadata to notebooks for search engine optimization. I created a Notion page on how to do so. Could you follow the steps and add metadata to this notebook?

https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e

e-g-melo:
Hi! I've reviewed the notebook and considered all your comments, including the metadata as well. Thanks @momchil-flex for including the Optimizer classes. In this version, the metallic reflector was removed, so the setup is the same one used to generate the structure used for the conference abstracts. The total optimization time was about 300 minutes this time at a cost of ~20 FlexCredits.

Thanks for all your comments!

momchil-flex:
Great. Maybe remove the `callback` from the optimizer as it's unused. Also, in the current state, if you load things from file you can't exactly continue the optimization in the same way as if you still have an `optimizer` instance stored in memory. This is because the gradient history stored in the `vt` and `mt` variables is lost. I'm not sure if anything should be done about this. One option is to pickle the optimizer, or to store the internal variables too in the json.

And I guess we need a "this notebook takes more than 10 credits" kind of statement on top? @tomflexcompute are we consistent about adding those?

e-g-melo:
OK. I've tried to pickle the optimizer state, but I got this error: PicklingError: Can't pickle : it's not the same object as pydantic.validators.callable_validator 

I will try the JSON option.

momchil-flex:
Don't spend too much time on it if it's not working out...

tomflexcompute:
> 

Yeah we should. We do have these warnings in recent notebooks too. Maybe for every new version release, we can use these warnings are indicators to not run these notebooks. Maybe only run large notebooks for major releases like 3.0, 4.0, etc. Otherwise notebook rerun can take a long time and the number of notebooks and the size of simulations are getting larger.

tomflexcompute:
The optimizer class indeed is very convenient. Don't we want to make it part of the adjoint plugin so others can use it directly?

e-g-melo:
Hi! I've included the optimizer state in JSON and put the warning message at the top of the notebook. Thanks!

tylerflex:
so are we waiting for anything else on this or can we merge?

PR #214, opened 2023-05-31
Added nonhermitian metagrating notebook

tomflexcompute:
Zongfu proposed a series where we recreate simulations in the latest and hottest research paperS and then invite the authors for an interview. If they agree, we can turn it into a blog and social media posts. This is the first one as a test to see how it goes.


momchil-flex:
Pretty cool. Just some minor comments:

- Is such a large size along z needed? Is it just to make the plots prettier / like in the paper?
- I was wondering why you do Ge using from_nk and chromium using a dispersive model (esp. given that it's a single-frequency simulation). Then I realized that Cr probably has Re(eps) < 0. I think you should explain that so that other people don't wonder too.
- `print(f'C_exc = {cal_C_exc(sim_data)}')` -> `print(f'C_exc = {cal_C_exc(sim_data):1.4f}')` to cut down the digits.
- when you plot the structure, maybe add a second plot zoomed around the grating? It's very hard to see right now. You can basically plot the same thing and then `ax.set_xlim` and `ax.set_ylim`.

But overall good to merge!

momchil-flex (approved):

caseyflex (approved):
Very nice demonstration! Just a couple small things to add to Momchil's comments.

- Don't forget to include the "Cr_Sytchkova.csv" file
- comment on warning in cell [3]

tomflexcompute:
Thanks @momchil-flex @caseyflex for reviewing!

PR #220, opened 2023-06-07
Gain medium in Dispersion notebook and update howdoi to include custom and gain medium

weiliangjin2021:
Update howdoi to include custom medium and gain medium


weiliangjin2021:
Note: for now running the notebook requires a special solver version that has been removed from the notebook.

tomflexcompute (commented):
Looks good! Is the incident pulse got amplified by the gain medium and then absorbed by PML so the simulation will not diverge? 

The previous simulation with the passive medium is also good. I wonder if we should keep that as it is and add a new section for the gain medium since most people are only interested in passive simulations. If they don't read carefully, when they see flux larger than one they might be misled to think that there is a normalization issue.

tylerflex (approved):
looks good, just a couple wording edits

weiliangjin2021:
> Looks good! Is the incident pulse got amplified by the gain medium and then absorbed by PML so the simulation will not diverge?

I think so.

>> The previous simulation with the passive medium is also good. I wonder if we should keep that as it is and add a new section for the gain medium since most people are only interested in passive simulations. If they don't read carefully, when they see flux larger than one they might be misled to think that there is a normalization issue.

> I agree with @tomflexcompute here, while it's probably ok overall, it might be slightly confusing if people just look at the figures without reading. One idea is maybe to add a thicker, lossy layer so that the final transmission is < 1?

In the revised notebook, I make the first layer lossy.

tomflexcompute (approved):
Looks good! Thank you for the revision.

PR #222, opened 2023-06-09
Tom/pic components

tomflexcompute:
We are creating code snippets to work as loadable pre-defined GUI script objects so users can use them to create common PIC components. This notebook can work as a tutorial to help users learn how to do the same in Python as many users don't know how to create curved structures like s-bends and circular bends.


lucas-flexcompute (commented):
This looks really useful! One question I have is whether the ring geometry will be correct with a sidewall angle. At the edges where top an bottom half-rings join, there will be gap, because the polyslab will shrink either above or below the reference plane, depending on the sign of the sidewall angle. I think you'd have to choose the reference plane so that the geometry only expands (correcting for the width and position accordingly) depending on that sign.

tomflexcompute:
Thanks @lucas-flexcompute for the great catch! Indeed that was an issue I didn't realize. The updated version should fix it. It seems correct to me now but could you double-check? 

One thing kind of puzzles me is that in the cross-section plots of the ring and directional coupler, there is a horizontal line in the middle. The cross-section plots of other components don't have this line. Do you know why that's the case?

lucas-flexcompute (commented):
That works for positive sidewall angles, but for negative (undercut etching) we should do the opposite: use the "bottom" as reference plane.

Regarding the split, could it be because the polyslab identifies an intersection of points and creates 2 sub-polyslabs? Do you see this split with sidewall angle set to zero?

tomflexcompute:
> That works for positive sidewall angles, but for negative (undercut etching) we should do the opposite: use the "bottom" as reference plane.
> 
> Regarding the split, could it be because the polyslab identifies an intersection of points and creates 2 sub-polyslabs? Do you see this split with sidewall angle set to zero?

Oh I didn't know it's possible to have a negative sidewall_angle. Is it common in real devices?

When sidewall angle is 0, I still see the splitting.

e-g-melo (approved):
Hi Tom! The notebook is great!

I have only two suggestions of structures that could be of interest:
1) A racetrack structure.
2) A square/triangular photonic crystal lattice.

weiliangjin2021:
> One thing kind of puzzles me is that in the cross-section plots of the ring and directional coupler, there is a horizontal line in the middle. The cross-section plots of other components don't have this line. Do you know why that's the case?

The horizontal line is artificial in visualization (the actual structure in simulation doesn't have the horizontal line). Let me think about how to remove the line.

PR #223, opened 2023-06-15
Updated the BilevelPSR notebook

tomflexcompute:
- Updated the BilevelPSR notebook due to some small sidewall_angle issue noticed by an user.

- Added cost estimation to a few notebooks to demonstrate how to estimate cost before running the simulation.

@tylerflex after the new estimate cost functions for job and batch are released, we can make another edit to demonstrate them in some other notebooks? I suppose there is no conflict with the current `estimate_cost()` in web?


tylerflex:
Yes, we can maybe call `estimate_cost()` from a `Batch`, maybe in the ParameterScan notebook, so users can see how to estimate the whole cost of the parameter scan before running. The new estimate_cost method is still awaiting merge into pre/2.3

PR #224, opened 2023-06-16
Added the exceptional coupling notebook

tomflexcompute:
- Added the exceptional coupling notebook
- Modified the image size in the Y branch notebook since it's a bit too large.


dbochkov-flexcompute (approved):
Looks great to me! Just minor comments:
- I think abbreviation `swg` was never defined, but it might be an issue only for people like me who are not too familiar with photonics fields
- Consider passing parameters to classes/functions always explicitly to increase readability, that is, for example, 
```
path_3 = gdstk.RobustPath(initial_point=(-1.2*L/2,0), width=[frac*p,frac*p,frac*p,frac*p,frac*p], offset=p, layer=1, datatype=0)
```
insted of
```
path_3 = gdstk.RobustPath((-1.2*L/2,0), [frac*p,frac*p,frac*p,frac*p,frac*p], p, layer=1, datatype=0)
```
- when defining mode monitors it seems like `L/2+R` and `(N+1)*p/2+w/2+R+lda0/4` can be defined as separate variables again to make it easier to follow

another question just of out curiosity: what makes this simulation to be so big? is it simply to demonstrate a very low crosstalk over a long waveguide, which wouldn't have been as impressive if the waveguide had been made shorter?

lucas-flexcompute (approved):
Looks good to me. Very interesting design.

tomflexcompute:
Thanks both @dbochkov-flexcompute @lucas-flexcompute for the prompt review.

@dbochkov-flexcompute thanks for the great suggestions. The notebook has been revised. And to answer your last question: yes you are totally correct, since the coupling between the waveguides with the cladding is somewhat small already, the coupling is noticeable only when the coupling regime is long.

PR #225, opened 2023-06-16
adds adjoint metalens notebook

tylerflex:
Requires JaxFieldData support from pre/2.3

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -88,6 +88,7 @@ Adjoint Plugin
     notebooks/AdjointPlugin_3_InverseDesign
     notebooks/AdjointPlugin_4_MultiObjective
     notebooks/AdjointPlugin_5_BoundaryGradients
+    notebooks/AdjointPlugin_7_Metalens
 
 Pre and Post Processing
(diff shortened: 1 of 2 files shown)


tylerflex:
Thanks Tom

> It looks like some cells are skipped. e.g. between [14] and [17] and between [19] and [21].

This is because I made some cosmetic changes and ended up clearing some outputs / removing cells, I will re-run.

> The final intensity plot looks a bit over-saturated. Might look nicer if we increase vmax a bit.

Sure, good point.

> In Conclusions, I think "or" should be "For". The markdown are cut off. I think you meant to say something like "the same principle can be applied to optimize a much larger metalens."

Yes, thanks, I fixed this.

> Above cell[10], the markdown says define a Simulation. I know that JaxSimulation is a subclass of Simulation, but would it be less confusing to readers if we just say and link to JaxSimulation directly?

Sure, good point

> Not necessarily an issue, but in [23] the apply_symmetry is set to False while it was True before. So the comparison looks a bit strange as one only contains a quarter while the other is the full structure.

Thanks I will fix this.

Re-running the notebook now but the next commit will address all of these!

PR #226, opened 2023-06-21
Added waveguide-to-ring coupling notebook

tomflexcompute:
- Added waveguide-to-ring coupling notebook
- Fixed a small typo in the y branch notebook.

@lucas-flexcompute this is essentially the replicate of your waveguide to ring coupling simulation in GUI. The goal is to demonstrate that using pml can lead to diverging. This is a common issue that our users experience. Sorry for having you as a reviewer frequently in the recent PRs!

@momchil-flex I added some notes on absorber but feel free to propose any additional information that can be added?


lucas-flexcompute (approved):
> Sorry for having you as a reviewer frequently in the recent PRs!

No problem at all! Here are a few comments:

In `In [8]`, the comment about the output mode monitor is duplicated. Maybe it could be specific to each monitor instance (through/drop output waveguides).

In theory, we could apply the absorbing boundary only at the upper y side, leaving a PML at the lower y, couldn't we? That would also be a good example of how to set a specific boundary condition at a single side. This is just a suggestion, though.

Before `In [13]`: "the simulation doesn't run into the **divergence** issue".

momchil-flex:
Is there a reason more than a half ring is simulated? Is it that the mode monitor results are more accurate when it is placed at a location where it is in-line with the ring radius, as opposed to placed closer to the bus waveguide with an angle setting, as illustrated in our modes in a ring example? Generally would be nice to take this opportunity to show how coupling to a bend with a very large radius can be analyzed while keeping a fairly small simulation domain.

Regarding the absorber comments, here's an edit suggestion:

> The adiabatic absorber is a multilayer system with gradually increasing conductivity. As briefly discussed above, the absorber usually has a larger undesired reflection compared to PML. In practice, this small difference rarely matters, but is important to understand for simulations that require high accuracy. There are two possible sources for the reflection from absorbers. The first, and more common one, is that the ramping up of the conductivity is not sufficiently slow, which can be remedied by increasing the number of absorber layers (40 by default). The second one is that the absorption is not high enough, such that the light reaches the PEC boundary at the end of the Absorber, travels back through it, and is still not fully attenuated before re-entering the simulation region. If this is the case, increasing the maximum conductivity (see the [API reference](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/e41b9db4c6161ba3b48851b8c068bbe6561dd455/docs/source/_autosummary/tidy3d.Absorber.html)) can help. In both cases, changing the order of the scaling of the conductivity (`sigma_order`) can also have an effect, but this is a more advanced setting that we typically do not recommend modifying.

tomflexcompute:
Thanks both. Using an angled source and reducing the simulation domain is a good idea. The results are largely unchanged. A FieldMonitor is also added to visualize the diverged field. Turns out the diverge occurs at the waveguide with the minimal transmission.

lucas-flexcompute (commented):
@weiliangjin2021 If you look at the arrows in the bent mode monitor, you'll see they are turning to the wrong side. I think you might have inverted the sign.

momchil-flex:
@weiliangjin2021 I think Lucas accidentally tagged you instead of Tom.

@lucas-flexcompute @tomflexcompute weird that the results don't change - maybe the bend radius was wrong in the first simulation already though, which should then explain that I believe.

tomflexcompute:
Sorry which part are you guys referring to? You mean the arrow direction is wrong in this?
image

lucas-flexcompute:
> Sorry which part are you guys referring to? You mean the arrow direction is wrong in this? image

Yes, the arrows are bending opposite to the ring curvature.

tomflexcompute:
The bend is very mild I didn't actually notice it 😅 . It seems like the radius should be negative in this case? I changed it to negative radius and the result is mostly the same but the arrow bending is correct now.

momchil-flex:
> Maybe at this radius, the bend mode is quite similar to the straight waveguide mode anyway, so it doesn't matter that much.

That would be my guess too.

Sorry, one last comment. This kinda looks like reflection to me. Can you see if it changes if you follow our advice and increase the absorber number of layers?

image

tomflexcompute:
Good point! 60 layers seem to do the trick.

momchil-flex:
Great, thanks!

PR #228, opened 2023-06-28
Updating custom field source with some extra usage info

momchil-flex:
@lucas-flexcompute could you have a quick look?


lucas-flexcompute (approved):
Looks good. The extra info should be very helpful!

PR #229, opened 2023-06-29
Added subpixel option in CustomMediumTutorial notebook

weiliangjin2021:
Revise the last cell since now we support subpixel averaging for custom medium.


momchil-flex:
Thanks. Seems like formatting is off.

image

weiliangjin2021:
You mean the extra line break between "To" and "apply". Somehow it's continuous locally.

weiliangjin2021:
Find out why. Let me fix it now

PR #232, opened 2023-07-06
Add nbdime to diff and merge notebooks (#722)

lucas-flexcompute:


tylerflex:
do you think this will be sufficient for improving our docs workflow? I dont know if I can test it out fully myself right now but it seems like adding it as a somewhat optional dependency like this is ideal and if it makes our lives easier I'm all for it.

lucas-flexcompute:
It certainly improves the workflow. I don't know if it solves all out pain points. Maybe @momchil-flex or @tomflexcompute can chime in (please take a look at https://github.com/flexcompute/tidy3d/issues/722)?

tomflexcompute:
@lucas-flexcompute sorry could you remind me what's the advantage of this implementation?

lucas-flexcompute:
Sure, for starters when I tried nbdev, it did not work. The generated notebook diff was incompatible with jupyter. It is also a more intrusive tool, whereas nbdime is more straightforward to use. Later I also found out that the Github diff viewer for notebooks is based on nbdime, so the rendering should be pretty much the same on Github and locally. The examples I created are on https://github.com/flexcompute/tidy3d/issues/722 if you want to take another look without installing it yourself.

tomflexcompute (approved):
I see. Thanks. Looks good to me.

PR #237, opened 2023-07-17
add adjoint utilities to notebooks

tylerflex:
Just a note, I kept the original functions in these notebooks but also added the convenience versions with a note. This way the user can either use our convenience functions or tweak the ones that were in there before. I could also be convinced to just remove the complex ones too.


lucas-flexcompute (approved):
I have only a few comments:

- I'm not a native speaker, but `filterer` and `projecter` sound very weird to me. Are those names intended?

- A better way to create the central coordinate for the boxes would be `coords_y = np.linspace(y0_min, y0_max, ny).tolist()` instead of `coords_y = [y0_min + index_y * size_box_y - 1e-5 for index_y in range(ny)]` in terms of floating point accuracy. Also why is there a 10⁻⁵ factor there? (This also appears in notebook number 6)

- "Fabricability Improvements" → Fabrication Constraints?

- Wouldn't it be better to also use the pre-made `AdamOptimizer` class in the notebboks?

e-g-melo:
Hi! 

Actually, I’ve started building the GC notebook using the adjoint mode converter notebook as example (https://docs.flexcompute.com/projects/tidy3d/en/stable/notebooks/AdjointPlugin3InverseDesign.html), so I copied the coords_x, coords_y code from there.

Regarding the initial GC structure, to start a new optimization you need to change the output_file variable. Otherwise, the optimization will continue from the last iteration. If the previous optimization was completed, the final result is loaded.

Emerson

lucas-flexcompute:
Juat a typo fix: "Ckeching".

tylerflex:
Thanks, I just fixed that typo, reran all notebooks (including the grating coupler) and made sure to reset all run functions to the server-side one (not local).

PR #239, opened 2023-07-19
Added a lectures section on the docs page

tomflexcompute:
To address https://github.com/flexcompute/tidy3d/issues/946, a "lectures" section is added. Under this section, currently there are only FDTD101 and Inverse design lectures (on-going). I assume we will have more in the future. What do you think @tylerflex ?

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -1,6 +1,15 @@
 Learning FDTD Through Examples
 ==============================
 
+Lecture Series
+--------------
+
+.. toctree::
+    :maxdepth: 1
+
+    fdtd101
+    inversedesign
+
 Examples of How to Use FDTD Features
 ------------------------------------
 
--- a/docs/source/fdtd101.rst
+++ b/docs/source/fdtd101.rst
@@ -1,0 +1,25 @@
+********
+FDTD 101
+********
+
+FDTD 101 is an introduction to the Finite-Difference Time-Domain Method for Electromagnetics. We will walk you through the basics of setting up and running electromagnetic simulations using the FDTD method. Through this course, you will gain a knowledge of the fundamental concepts behind electromagnetic simulation as well as many advanced topics worth considering when you set up your simulations.
+
+`Lecture 1: Introduction to FDTD Simulation <https://www.flexcompute.com/fdtd101/Lecture-1-Introduction-to-FDTD-Simulation/>`_
+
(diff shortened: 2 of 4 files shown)


docs/source/examples.rst
```diff
@@ -1,6 +1,15 @@
 Learning FDTD Through Examples
 ==============================
 
+Lectures
```
tylerflex, line 4: Maybe "Lecture Series"?

docs/source/fdtd101.rst
```diff
@@ -0,0 +1,29 @@
...
+
+`Lecture 7: Time step size and CFL condition in FDTD <https://www.flexcompute.com/fdtd101/Lecture-7-Time-step-size-and-CFL-condition-in-FDTD/>`_
+
+`Lecture 8: Numerical dispersion in FDTD <https://www.flexcompute.com/fdtd101/Lecture-8-Numerical-dispersion-in-FDTD/>`_
+
+`Lecture 9: Dielectric constant assignment on Yee grids <https://www.flexcompute.com/fdtd101/Lecture-9-Dielectric-constant-assignment-on-Yee-grids/>`_
+
+`Lecture 10: Introduction to subpixel averaging <https://www.flexcompute.com/fdtd101/Lecture-10-Introduction-to-subpixel-averaging/>`_
+
+`Lecture 11: Introduction to Inverse Design In Photonics <https://www.flexcompute.com/fdtd101/Lecture-11-Introduction-to-Inverse-Design-In-Photonics/>`_
```
tylerflex, line 27: Perhaps leave these out of here as they are in inversedesign.rst

tylerflex (approved):
looks good, thanks for adding them @tomflexcompute ! just a couple minor comments then go ahead and merge.

docs/source/fdtd101.rst
tomflexcompute: Oh just realized they are the same lectures. Will change that.

tomflexcompute: Actually since all lectures are on the FDTD101 page, should we just have one lecture here too without separating out the inverse design? Are all future lectures to be added to the same page?

tylerflex: either way. they will be split eventually. we just didn't have time yet to make a whole new website for inv des

tomflexcompute: I see. Then in that case I will have them separated here and then update the links once they have a separate website.

PR #240, opened 2023-07-19
Updated Fitting notebook to showcase new FastDispersionFitter

caseyflex:


momchil-flex:
Great. I would suggest removing the DispersionFitter completely. Too often we get divergence because of active materials.

On the other hand, how about expanding the fast fitter section a bit, specifically with more tips or illustrations on using some of the advanced parameters? If you can think of something that may be commonly of use (e.g. when fitting metals? or some difficult models)

weiliangjin2021 (approved):
Looks very good to me!

tylerflex:
@momchil-flex feel free to merge this if your comments are addressed.

momchil-flex:
Looks great now, I think this will be super useful going forward.

PR #241, opened 2023-07-20
Alec/photonic crystals

alec-flexcompute:


tomflexcompute (commented):
Thanks @alec-flexcompute . I think the functions all work well. This will be very helpful for the users. In terms of formatting, I list a few comments:

Seems like your notebook is created in colab so there is a colab icon on top. Not sure how it works but could you use the regular jupyter notebook? 

The "!pip ..." and "web.configure.." need to be deleted in the first cell, as well as the "web.test()" on the second cell.

Could you write some general intro about this notebook as a markdown at the beginning of the notebook? Like what's the purpose of this notebook (provide grab-and-go functions to help users define their structures quickly) and what is provided (common periodic photonic structures like hexagonal lattice hole slab, woodpile, ...).

Instead of inserting a screenshot image taken from the web interface, can we do an example case after each function and use Structure.plot() to visualize it? It will create a plot like you did for the L cavity for example.

alec-flexcompute:
Hi Tom,

Sure I can install regular jupyter notebook, except it might take a little time to get used to it. Is there a reason we don't want the colab icon on top?

For structure.plot, for some of these structures it's hard to grasp the full shape using 2d plots. Is there a plotting function that gives a 3d angle?

tylerflex:
You can use this for 3D plotting: https://docs.flexcompute.com/projects/tidy3d/en/stable/_autosummary/tidy3d.Simulation.html?highlight=plot_3d#tidy3d.Simulation.plot_3d

And I think it would be good to set up local jupyter notebook (it's relatively straightforward) and run with that just for consistency and so all the notebooks look the same.

alec-flexcompute:
Uploaded jupyter-notebook with updated style/display options, again labeled "PhotonicCrystals.ipynb"

tomflexcompute:
@alec-flexcompute I pushed a series of formatting changes to be more consistent with other notebooks. Please take a look and see if they are working correctly and make sense. 

In the grating case, the definition of the axis is slightly confusing to me. For example, for axis=2, i.e. z axis, intuitively I would imagine the grating to be facing up instead of extruded in the z axis. Do you think you can fix this?

alec-flexcompute:
Yeah for sure

alec-flexcompute:
Just added it

tomflexcompute:
Thanks. I'll check a few more things and merge the PR to develop later.

tomflexcompute:
@alec-flexcompute I added a few small formatting changes and metadata. It looks like it's ready to be merged. Could you take a last look to see if everything looks fine? If so, feel free to square all commits and rebase/merge into develop.

alec-flexcompute:
For sure - this is the PhotonicCrystalsComponents notebook, right? Should I get rid of the PhotonicCrystals notebook, as it's a duplicate but with the old formatting?

tomflexcompute:
> For sure - this is the PhotonicCrystalsComponents notebook, right? Should I get rid of the PhotonicCrystals notebook, as it's a duplicate but with the old formatting?

Yes correct. Please delete the old notebook.

alec-flexcompute:
Deleted old notebook and everything looks good. Squashed commits and pushed.

PR #252, opened 2023-08-08
Notebooks updates for colocate refactor

momchil-flex:
I wanted to make a PR that shows just the changes that I need to make to the notebooks for the monitor refactor. You can explore the [docs build](https://docs.flexcompute.com/projects/tidy3d/en/momchil-colocate/examples.html) in which I reran all notebooks using that solver and which I used to figure out what changes need to be made. To have an easy to review PR, I've only incorporated those changes here, without actually rerunning the notebooks. Notably:

- `colocate = False` explicitly set in a few places in monitors that would be used to do user-side field projections. This is to avoid the double colocation warning, although even the results shouldn't be affected much even if left unset.
- Fixed the last example in the field projection notebook to have the projection monitor slightly away from the source rather than sitting on top of it.
- `colocate = False` also explicitly set in the custom source notebook, with some explanation as to why.
- Some explanation about `at_cetners` and `at_boundaries`.

I was thinking of also adding a section in the FAQ about Yee grid, primal grid, dual grid, boundaries/centers, what it all means, where do fields live by default, etc. I even worked on a figure. But the more I thought about it the more it seemed hard to explain simply, so I'm wondering if there's even a point in discussing this in the FAQ. What do you think?

yee_grid
e a


docs/source/notebooks/Metalens.ipynb
shashwat-sh, line 1: I think the first simulation must have run on a different node / cell than before, so some of the times we hard-coded in the text are a bit off. Probably not related to this PR specifically, but just caught my eye:
- Below cell 11, instead of "1 minute" maybe we can generically say "order of minutes" or "a few minutes".
- Below cell 17, instead of "factor of 3" we could say "several times faster" or something.
These could guard us against future changes in run time.

shashwat-sh (approved):
Looks good to me, one minor comment for the metalens.

Regarding Yee grid explanations, generally I like the idea of educating users and being transparent about the fundamentals of our approach, but to avoid the complications you mentioned, maybe you could instead link to existing free resources on this, such as our FDTD lectures?

I think you could still include the figures you made (maybe add x, y, z axes just for clarity, even though it doesn't really matter) because they make it very clear that we don't use the same Yee scheme as the Taflove / Hagness book, which could be important to know for some users.

tylerflex:
Yea I agree with @shashwat-sh about the extra yee discussion. I feel it would be super useful to have **some** discussion and to include the figure. I guess I'd include as much as you can and once it gets too complicated I guess link to something else?

tylerflex (approved):

tylerflex:
Notebook additions and discussions look good to me!

docs/source/notebooks/Metalens.ipynb
momchil-flex: Yeah generally we should avoid hard-coding times. This has been an issue before with that notebook too. Honestly I'm down to just remove the discussion under cell 11. Just remove the whole sentence, "As we can see from the logs, the total time to solve this problem (not including data transfer and pre/post-processing) was on the order of a minute!" and leave the "For reference..." one under it. @tylerflex ?

Also under cell 17, maybe just avoid discussing speedup altogether? We mention 10x less data and we could also mention that it's cheaper.

momchil-flex:
Alright I also added a yee grid discussion and the figure, have a look!

docs/source/notebooks/Metalens.ipynb
shashwat-sh: Sounds good to me

shashwat-sh:
I think the Yee grid explanation reads very nicely, should be pretty easy to follow for those unfamiliar with it.

tylerflex (approved):
Yea looks good, very clear. thanks

docs/source/notebooks/Metalens.ipynb
tylerflex: yea let's be more vague about times I agree.

PR #254, opened 2023-08-10
add waveguide bend

tylerflex:
Also adds the inverse taper after re-run with improved numerically stable penalty function, demonstrating much faster convergence.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -93,6 +93,7 @@ FDTD Adjoint Optimization
     notebooks/AdjointPlugin5BoundaryGradients
     notebooks/AdjointPlugin6GratingCoupler
     notebooks/AdjointPlugin7Metalens
+    notebooks/AdjointPlugin8WaveguideBend
 
 Tidy3D Case Studies
 -------------------
--- a/docs/source/notebooks/metadata/metadata.txt
+++ b/docs/source/notebooks/metadata/metadata.txt
@@ -53,7 +53,13 @@
     "keywords": "inverse design, metalens, design optimization, adjoint, Tidy3D, FDTD",
     "feature_image": "N/A"
   },
-  "AndersonLocalization.ipynb": {
+  "AdjointPlugin8WaveguideBend.ipynb": {
+    "title": "Adjoint Optimization of a Metalens in Tidy3D | Flexcompute",
+    "description": "This notebook demonstrates the adjoint optimization of a waveguide bend in Tidy3D using the adjoint plugin.",
+    "keywords": "inverse design, bend, design optimization, adjoint, Tidy3D, FDTD",
+    "feature_image": "N/A"
+  },
+    "AndersonLocalization.ipynb": {
     "title": "Anderson Localization of EM Waves in 3D | Flexcompute",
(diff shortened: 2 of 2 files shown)


tomflexcompute (changes requested):
Great demonstration! A few comments:

Probably better to replace "0th order mode" with the more commonly used "TE0 mode"?

Links to other notebooks are generally encouraged to use the pages at https://www.flexcompute.com/tidy3d/examples/ instead of the docs site. Could you change them? 

geo_sub is defined from -10000 to 0 in z, but the waveguide structures are from -h/2 to h/2. They will overlap in this case, which is shown in the plotted simulation cross-section. Shouldn't the substrate be -10000 to -h/2?

The Ex plot in the second row has a different colormap, which looks a bit strange.

Comment in cell[4], "refractive index of waveguide and substrate" -> "refractive indices"?

Above cell[5], I think either "simulation size" or "simulation's size"?

Above cell[10], "the simulation domain son we need to" -> "the simulation domain so we need to".

Above cell[10], "accross" -> "across".

Above cell[20], "paraneters" -> "parameters".

Above cell[22], "innacurate" -> "inaccurate".

Comment in cell[24], "approporiate" -> "appropriate".

Above cell[26] and in the comment of [29], "funciton" -> "function".

tomflexcompute:
Btw the refractive index of SiN is actually around 2 (1.974).

tylerflex:
> Probably better to replace "0th order mode" with the more commonly used "TE0 mode"?

fixed, thanks.

> Links to other notebooks are generally encouraged to use the pages at https://www.flexcompute.com/tidy3d/examples/ instead of the docs site. Could you change them? 

fixed

> geo_sub is defined from -10000 to 0 in z, but the waveguide structures are from -h/2 to h/2. They will overlap in this case, which is shown in the plotted simulation cross-section. Shouldn't the substrate be -10000 to -h/2?

oops, good catch. thanks, I fixed this.

> The Ex plot in the second row has a different colormap, which looks a bit strange.

hardcoded them all to "RdBu" 

> Comment in cell[4], "refractive index of waveguide and substrate" -> "refractive indices"?
> Above cell[5], I think either "simulation size" or "simulation's size"?
> Above cell[10], "the simulation domain son we need to" -> "the simulation domain so we need to".
> Above cell[10], "accross" -> "across".
> Above cell[20], "paraneters" -> "parameters".
> Above cell[22], "innacurate" -> "inaccurate".
> Comment in cell[24], "approporiate" -> "appropriate".
> Above cell[26] and in the comment of [29], "funciton" -> "function".

Thanks, I made all of these changes too! I guess my jupyter lab spell checker isn't working for some reason.

I will re-run the notebook now and push the final version.

tylerflex:
I also changed the waveguide refractive index to 2.0

tomflexcompute:
Looks good! Thanks for the modifications. Did you add metadata? If not, maybe I can do it next time with some other new notebooks that are about to be released. 

Usually transmission is plotted as a function of wavelength instead of frequency but it's fine for now. We can change it and rerun in the future.

tomflexcompute (approved):
Thanks.

PR #255, opened 2023-08-17
Use hole support in notebooks

lucas-flexcompute:


tomflexcompute:
Thanks for adapting the notebooks. Are we removing PolySlab.from_gds in the future?

lucas-flexcompute:
I'm not sure. `Geometry.from_gds` will return a `PolySlab` or a `ComplexPolySlabBase` if the geometry is appropriate, but I think removing `PolySlab.from_gds` might be a big breaking change.

BTW, I did not change the photonic crystal functions notebook, although the hole versions could be similar to the photonic crystal waveguide notebook. I though having the functions as they are is not really any problem.

tomflexcompute (approved):
Thanks @lucas-flexcompute . All look good. There are a few other notebooks that can be adapted to this new addition, but we can work on that later on since the old implementation will still work fine.

momchil-flex:
Alright I'll merge this in so that I use these notebooks in the rerun.

PR #257, opened 2023-08-18
fix adjoint grating coupler notebook

tylerflex:
the `beta_i` used in the final simulation / plots never got set if the iteration loop is skipped.

This PR sets it before the loop to be the final value based on the scheduling. 

Results are equivalent.


tylerflex:
Note: the iteration printing is removed with this commit since I loaded it from file. (not sure how to keep that).

Instead, it should suffice to just insert the line

```
beta_i = init_beta + total_iter * del_beta # final beta, if the below loop is skipped
```

at the beginning of cell [15] before the optimization loop is entered.

PR #258, opened 2023-08-18
Added feature images to notebooks that are missing it

tomflexcompute:
- added feature images to 4 notebooks that are missing it.
- added a feature image to focusing apodized grating coupler notebook to make it look more consistent with the rest. 

MC is working on the [learning center](https://develop.d3nzcgsw5oo0x1.amplifyapp.com/tidy3d/learning-center/example-library/) where the example library section will automatically use feature images from the repo. Currently there are 4 notebooks without a feature image and the focusing apodized grating coupler has a slightly different styled image.

@momchil-flex this PR shouldn't cause any issue in terms of the 2.4 release?


momchil-flex:
> @momchil-flex this PR shouldn't cause any issue in terms of the 2.4 release?

Actually it can be a bit tricky as I'm re-running all notebooks, but hopefully merge with -X theirs will work again...

I guess let's get it in and deal with it.

PR #260, opened 2023-08-22
Added import file mapping json

tomflexcompute:
- Added import file mapping json. This file shows which notebook imports which files from the misc folder. This is for MC to use in our hosted Python notebook environment. When a user creates a task from one of our example notebooks, the associated imported files from the misc will be made available so the user can run the notebook without any error. 

- Also added a quick demonstration of loading simulation data from hdf5 as requested by Xin since many users don't seem to know about it.


tomflexcompute:
- Also added plot_3d() to a few suitable notebooks to demonstrate 3D plotting. Planned to do this a while ago but didn't get to it until now.

tylerflex (approved):

PR #261, opened 2023-08-28
Added MoS2 monolayer waveguide notebook and associated files

tomflexcompute:
This is the monolayer MoS2 waveguide notebook inspired by a recent Science paper. There will be an author interview after this notebook is published on our website.


e-g-melo (approved):
Hi Tom! Very nice the notebook! In addition to the novelty presented by the paper, it will be useful to illustrate the usage of the new FastDispersionFitter and the Medium2D.

I have only some minor comments on it:
1- add a space between "Science381"
2 - "Our simulations confirmed that The real part" -> "Our simulations confirmed that the real part"
3 - I suggest to mention something like "We have used the FastDispersionFitter and Medium2D features to model the WoS2 material" at introduction, so users can easily find this notebook when looking for how to use these features.
4 - Is not `weights=(1,1)` in cell [2] the default value?
5 - Is the only purpose of coupling the light to free-space at waveguide end to show it has been guided through MoS2? Or is that an edge coupler?
6 - I don't have access to the paper. Is it only based on simulations or it also include experimental measurements?
7 - Is there any comments or comparisons to other materials regarding waveguide propagation losses?

tomflexcompute:
> 

Thanks @e-g-melo for the quick review and great suggestions! 

1. Fixed.
2. Fixed.
3. Added in the introduction. 
4. Indeed `weights=(1,1)` is the default value. Here it's more for showing that it can be tuned but since the n and k are the same order already, we don't need to manipulate the weights. I've added comments on this to make it more clear. 
5. In the paper I think they only demonstrate the guiding. It has not been demonstrated to be integrated with other photonic devices, which might be their next research direction. 
6. The paper is primarily an experimental paper actually. They did measure the propagation loss, which matches up pretty nicely with our k_eff.
7. Comments on using other 2D materials as waveguides and potentially more sophisticated devices have been added at the end of the notebook.

momchil-flex:
Should this be merged in 2.4 rather than develop?

tomflexcompute:
> Should this be merged in 2.4 rather than develop?

It seems to me that it's ok to merge to develop (and then to latest) since FastDispersionFitter is already available in the latest version?

momchil-flex:
No, fast dispersion fitter is only available in the 2.4 pre-releases. `latest` points to our latest docs referring to our latest official release (2.3.3) while `stable` points to the docs that were built whenever that release was published. Pre-release docs are built from the pre branches.

tomflexcompute:
> No, fast dispersion fitter is only available in the 2.4 pre-releases. `latest` points to our latest docs referring to our latest official release (2.3.3) while `stable` points to the docs that were built whenever that release was published. Pre-release docs are built from the pre branches.

Maybe I didn't understand it correctly. In this case, the latest is actually behind 2.4.0rc2? Then indeed it should be merged into 2.4.

momchil-flex:
Basically:
- We release e.g. 2.3.3. We make a v2.3.3 tag on tidy3d-docs. This is what `stable` will point to until there's a new tag that's not a pre-release, e.g. 2.4.0. We also merge this in `latest`.
- Before our next stable release, we may want to keep making changes to the 2.3.3 docs, like typos, new notebooks, ... These get merged in develop and from there in `latest` branch, where they automatically appear under the `latest` tag in the docs, which is the default landing page.
- Meanwhile we are working on pre-releases. We release 2.4.0rc1. This gets its own tag and can only be accessed through that tag in the docs page. 2.4 doesn't go in latest until it is official and not pre-release.

Right @tylerflex ?

P. S. Now the PR needs rebase. You could try `git rebase -Xours pre/2.4`. `Xours` in a rebase means resolve conflicts in favor of `pre/2.4`.

tylerflex:
yea that seems right to me based on what we discussed.

tomflexcompute:
> Basically:
> 
> * We release e.g. 2.3.3. We make a v2.3.3 tag on tidy3d-docs. This is what `stable` will point to until there's a new tag that's not a pre-release, e.g. 2.4.0. We also merge this in `latest`.
> * Before our next stable release, we may want to keep making changes to the 2.3.3 docs, like typos, new notebooks, ... These get merged in develop and from there in `latest` branch, where they automatically appear under the `latest` tag in the docs, which is the default landing page.
> * Meanwhile we are working on pre-releases. We release 2.4.0rc1. This gets its own tag and can only be accessed through that tag in the docs page. 2.4 doesn't go in latest until it is official and not pre-release.
> 
> Right @tylerflex ?
> 
> P. S. Now the PR needs rebase. You could try `git rebase -Xours pre/2.4`. `Xours` in a rebase means resolve conflicts in favor of `pre/2.4`.

ok sounds good. Should be fine now.

momchil-flex:
Something is still wrong, it's trying to merge 15 commits and too many files.

If rebasing to 2.4 doesn't work well, you might have to branch off of 2.4 and checkout only the files you have modified...

tomflexcompute:
> 

I think it's just bringing some recent differences from develop to pre2.4. But ok let me branch out pre2.4 directly to avoid any issues.

PR #266, opened 2023-09-07
added bistability notebook

alec-flexcompute:


tomflexcompute (changes requested):
Thanks @alec-flexcompute for the great case study. As discussed, the result is not identical to the reference but I won't worry about it too much. Below are my comments. Please let me know if you have any questions or find anything unclear.

1. What is "View this project in Tidy3D Web App" leading to? Is it supposed to link to our web GUI? Currently it's linking to another notebook page, which is a bit confusing to me.
2. When linking to the publication, ideally we link to the DOI link, which in this specific case is "https://doi.org/10.1063/1.1615835". This paper is not open-access so viewers need to purchase the pdf to actually read it. The same for the second paper by Soljacic et. al.
3. "Tidy3d" -> "Tidy3D". 
4. fwidth = freqs[-1]-freqs[0] will give a source spectrum that is much wider than necessary. Although it might not hurt, usually we keep it 0.5*(freqs[-1]-freqs[0]) or even narrower. This also prevents the excitation of other resonances in the system.  
5. There are a few RunTimeWarnings. Do we know why they show up?
6. Could you briefly comment on why absorber instead of PML is used? Essentially the evanesent field around the rods will leak into PML and cause the simulation to diverge.
7. In the comment you mentioned "We plot the flux through the waveguide vs frequency, and then visualize the field at this frequency". However the field distribution in the linear sim is not actually plotted. 
8. Might be nice to add a print statement after the calculation of kappa to display its value. 
9. Above cell[14], "Then then" -> "Then".
10. Using the original figure from the paper requires reproduction permission from the publisher. Let's avoid showing the figure and just state that it matches Figure 4 from Yanik et. al.
11. Link CustomSourceTime in the commend to its API reference page. 
12. The nonlinear_source is the same as the linear source but with a different source_time? If so, using the copy method and only updating the source_time is more elegant. 
13. Header capitalization is not consistent throughout the notebook. Please double check.  
14. The nonlinear source spectrum has some negative amplitudes. Is it some numerical artifact? 
15. The Ez plots are a bit saturated. We can manually set vmin and vmax so it looks a bit better. Usually I set vmax (vmin) to slightly below the actual maximum (minimum) field value. 
16. In Introduction, "We base this example off of the paper" -> "We base this example of the paper".
17. In Computing kappa, "permmittivity" -> "permittivity".
18.  In Plot Transmission and Analytic Solution, "calcuation" -> "calculation".
19. You also need to add this notebook to the examples.rst file so when the docs side is built, this notebook will show up.

alec-flexcompute:
Thanks Tom for your comments. I've implemented these changes. A couple things:
5. The runtime warnings are: "WARNING: Simulation has 3.47e+06 time steps. The 'run_time' may be unnecessarily large, unless there are very long-lived resonances."
11. When writing this I could not find a link to a CustomSourceTime description. When writing the code I had to refer to the actual files to see how it worked - we should add an API description?
14. This should just be a nonlinear artefact, as this is just plotting the sourcetime and doesn't cause any problems
16. Is "We base this example of the paper..." grammatical? I'm pretty sure it should be "We base this example off of the paper..."

tomflexcompute:
> 

Thanks Alec. The warnings that I'm more concerned about are
image

The 2.4 docs will probably need to be built with the 2.4 release. I think you can probably just link to like "./_autosummary/tidy3d.CustomSourceTime.html" but if you want to be more conservative, we can leave the link now and add it later on after 2.4 is released and docs are built.

Does "We base this example off of the paper..." sound correct? If you are sure about it, let's just go with it. I clearly don't have much expertise in English :joy: I put it in Grammarly and it was picked up as an error.

caseyflex:
Thanks, this is very exciting and coming together nicely! Here are some initial comments.

- In cell [4], can we eliminate the need to already know fres, for example by using a broader frequency range pulse?
- In cell [10], you plot the transmission spectrum of the waveguide + linear cavity system. It would be good to go ahead and run the waveguide-only simulation here, so we can normalize the transmission spectrum. This can replace the "Waveguide Normalization" section you have later on, since we will already know the waveguide transmission at the detuned frequency.
- Also in cell [10], I thought we had a cleaner transmission spectrum previously with less ringing and a clearer Lorentzian lineshape.
- In many plots like cell [8], you truncate the simulation to hide the boundary. It would be nice to show the full simulation setup in these plots.
- In cell [14] when computing kappa, can you clean up and comment this code a bit? For example, `y//2 + 6` is a bit opaque
- In cell [15], we should reference Boyd's Nonlinear Optics for the equation for chi3
- In cell [22], I don't think we need to plot the source spectrum, as it is essentially at a single frequency except for finite run_time artifacts
- In cell [23], do we need interval? Can we just have start = stop for the FieldTimeMonitors?
- In cell [37] and [38], we have some hardcoded numbers, can we make these variables or computed quantities?
- Cell [31] I think is unnecessary / not the best way to do this. To reiterate an earlier comment, you can just do linear regime pulse simulations around cell [10] for waveguide-only system, and use the transmission at the detuned frequency to normalize the power.
- In cell [32], I see that you are only using the computed transmission to normalize for the coupled-mode theory. It should actually be used around cell [16], as every value P/P_0 is using the transmitted P in the waveguide-only system, while the input power curve needs to be slightly greater to compensate for incomplete transmission.

caseyflex (changes requested):
See above comment for requested changes.

alec-flexcompute:
Hi Casey, a couple things:
Comment 1: I did do that initially, but this greatly increases the flexcredit cost, so I thought to avoid it. If you want I can add more frequencies to record though
Comment 2: when plotting the transmission spectrum, I'm using a different source than the ramp-up-pulse source, as the latter takes a lot longer to run. Would we run into problems here if I normalize the transmission spectrum?
Comment 4: I truncated the plots of the simulations because they look weird without any bounds. Do you know how I could clean this up in a different way?
image
For comment 9, this is why I was asking about plotting according to a 'nearest' function, and we decided on just using hard-coded values for now, although I could try some function within xarray
Also, for comment 3, the cleaner result had the source and monitors closer to the boundary, which we had decided against. LMK if you want me to move them back, though

caseyflex:
> Hi Casey, a couple things: Comment 1: I did do that initially, but this greatly increases the flexcredit cost, so I thought to avoid it. If you want I can add more frequencies to record though Comment 2: when plotting the transmission spectrum, I'm using a different source than the ramp-up-pulse source, as the latter takes a lot longer to run. Would we run into problems here if I normalize the transmission spectrum? Comment 4: I truncated the plots of the simulations because they look weird without any bounds. Do you know how I could clean this up in a different way? image For comment 9, this is why I was asking about plotting according to a 'nearest' function, and we decided on just using hard-coded values for now, although I could try some function within xarray Also, for comment 3, the cleaner result had the source and monitors closer to the boundary, which we had decided against. LMK if you want me to move them back, though

1. Why does a broader frequency range increase the flex unit cost? Is it due to Nyquist sampling? If so, maybe you can still broaden it sufficiently but not too far without increasing the cost unreasonably.
2. When you're in the linear regime, the detailed pulse shape isn't so important. That's why we usually use GaussianPulse in these simulations.
4. Why does the crystal extend on the left side? Seems like it should just be air here. And the same absorber size on the left and the right should be fine.
9. When I set start = stop in the FieldTimeMonitor, I don't need to even specify a time. Something like `sim_data.plot_field(field_monitor_name='field', field_name='Ez', val='abs', z=0, ax=ax)` works fine for me, where 'field' is a FieldTimeMonitor.
3. Do you know why this would be? I think their current location is fine. Not crucial to remove the ringing, mostly cosmetic.

alec-flexcompute:
1. By broader frequency range, do you mean that the monitor records?

3. The crystal was extended beyond the boundary, and to add air, I added a block with the air medium on the side.

5. I'm not sure - I had wondered about this when I first got the results. It also might have something to do with the source not being in the crystal in the first place.

caseyflex:
> 1. By broader frequency range, do you mean that the monitor records?
> 2. The crystal was extended beyond the boundary, and to add air, I added a block with the air medium on the side.
> 3. I'm not sure - I had wondered about this when I first got the results. It also might have something to do with the source not being in the crystal in the first place.

1. Yeah, I mean for both the monitors and source pulse fwidth. If the run_time stays the same and you play with shutoff, then you should get the same spectral resolution.
2. Ok, let's get rid of it for clarity.
3. It shouldn't have much to do with the source not being in the crystal. Usually, it would relate to insufficient run_time / too high shutoff threshold.

tomflexcompute:
I think a more suitable category for this notebook would be "Photonic Crystals" instead of "Nanophotonics"?

Also I forgot to mention that could you update the metadata.txt file under the metadata folder?

alec-flexcompute:
So normalizing by the transmission from the waveguide-only simulation:

power_curve()/np.sqrt(detuned_transmission)

results in a very different curve from that of the paper:

alec-flexcompute:
image

caseyflex:
You obtain different results with these two strategies, even though both
are based on a linear simulation?

caseyflex:
Maybe you need to normalize the waveguide transmission itself based on the
GaussianPulse source spectrum?

alec-flexcompute:
So the curve above is obtained from the following: taking the initial source used to calculate the cavity resonant frequency, I use that as the source for just the waveguide. I obtain the transmission flux at the detuned frequency, and this flux is the "detuned_transmission" variable above - the square root of which I divide the power values with to get the blue curve.

Is this what you meant by your comment #2 above?

caseyflex:
I’m not sure why you aren’t getting the same result with this approach. I
was just suggesting checking sim.normalize_index, but this is probably not
the issue.

alec-flexcompute:
Should I do the previous approach then?

caseyflex:
No, I think you should try to figure out what’s wrong with the new setup.

caseyflex:
I can also take a look if you share the new setup notebook

caseyflex:
Hi @alec-flexcompute,

I played with the notebook a bit. When I use `values=power_curve(times*gamma)/np.sqrt(detuned_transmission)/np.sqrt(2)` for the `CustomSourceTime` and I set `power_input = scipy.interpolate.interp1d(times_nonlinear, np.sqrt(2)*np.abs(source_time.amp_time(times_nonlinear))*np.sqrt(detuned_transmission))`, I get the following quite satisfactory curve. We need to multiply the amplitude by sqrt(2) or the power by 2 to relate the peak power to the cycle average.

image

alec-flexcompute:
Thanks Casey for this key fix! I got the same figure. Added an extra note, changed to photonic crystals, and pushed. LMK if there's any other improvements you'd like

caseyflex:
Hi Alec, this is looking great! A few more comments.

1. Cell [6] and [8], we can lower (or set to None so we have control via run_time) the shutoff to reduce or eliminate the ringing.
2. Cell [6], can we show the entire simulation without any limits? You can use plt.figure(figsize=(10, 10)) or something to make the figure larger.
3. Cell [5], we might be able to reduce the number of absorber layers in the x direction.
4. Cell [5], I think we should set `subpixel=False` to get better agreement with the paper.
5. Cell [7], it seems like we are downloading 500 MB of data, which is not absurd but perhaps we can reduce this. For example, instead of using the FieldMonitor at so many frequencies to find the resonant mode profile, perhaps we could use a single FieldTimeMonitor at like `t=20/fwidth`.
6. Cell [26] and [27], can we show the entire simulation?
7. Cell [20], if you want to show the spectrum, you can pass in more times to get rid of the negative peak artifacts.
8. Cell [26] and [27], we should use the same vmax scale. It is also weird to see lower values before and after the cavity in the second figure. Can you figure out what is going on here, and fix it?
9. Cell [20], it's a bit unclear to say the cycle average at 1/2 the peak power. It's just that the cycle average power is equal to 1/2 the peak power.

alec-flexcompute:
So when plotting the simulation, automatically the plot looks like this:
image

alec-flexcompute:
For comment 6, how do these field profiles look? (with the same limits)
image
image

tomflexcompute:
Thanks @alec-flexcompute for all the great efforts. Looks very good. 

I think we generally try to get rid of these console outputs with plt.show()
image

We need to merge into develop at this point so you probably need to rebase there.

alec-flexcompute:
Sounds good. I'll add that and wait for Casey's comments before pushing again.

caseyflex:
Thanks Alec, the notebook is looking really nice. Your efforts on the DBR seem to have paid off! Some minor comments:

1. After Cell [1]: .2a instead of 2a. Also, comment that for simplicity we use the DBR parameters from the paper.
2. Cell [2]: `dir` is reserved in python, best to use a different variable name here
3. Cell [3]: `semi_major` and `semi_minor` are technically just `major` and `minor`
4. Cell [4]: gridSpecAuto not used
5. After Cell [4]: change link to DOI. 
6. Cell [5]: rename absorber_y since it's not an absorber. I'd also prefer periodic to PEC here.
4. Cell [4]: gridSpecAuto not used flux_interval. 
5. Cell [5]: In principle, we can use FieldTimeMonitor instead of these evenly spaced in frequency FieldMonitors. After t_start, ideally only the resonant field should be present, so this would simplify the setup and remove some error.
7. Cell [5]: I'd like to set `shutoff=False` as the field decay isn't really relevant, since fwidth is much broader than our resonance.  Then, `run_time=10000/freq0` or something should still work fine.
8. Before Cell [8]: I'd rephrase this as "the transmission vanishes on resonance" or something
9. Cell [10]: can you add a note here, that the result from the ResonanceFinder agrees with what we see in the transmission spectrum above?
10. Cell [12]: only need to use Ez, as Ex and Ey vanish due to symmetry. Also, can we plot the modal profile?
11. Cell [22]: don't need interval
12. Cell [26]: could probably reduce saturation threshold to make this clearer
13. Cell [29]: power_input line is too long, I can't see the end in my browser
14. Cell [29]: comment on possible sources of error. DBR parameters, kappa estimation, grid, ...

A general comment: whenever I have a monitor, I like to somehow plot or visualize the data recorded by that monitor. For example, after cell 8, we could also show the FieldTimeMonitor data, and the field profiles. This is good practice to check that everything is working correctly.

caseyflex (approved):
Great, looking good! 

I was looking at the last figure again and I see that the discrepancy already exists in the linear regime. This surely means that the error is only due to the DBR parameters, so maybe we should get rid of the other reasons to avoid confusion. 

Also, I noticed that the T function in the last cell is not used, so maybe we can delete it.

Then, I think it's good to go. @tomflexcompute can approve the PR too if he agrees.

tomflexcompute:
Thanks @alec-flexcompute and @caseyflex for the continuous efforts. It looks very nice and ready to go.

A side note: we were discussing doing integrals directly using xarray. See the last comment from Tyler [here](https://github.com/flexcompute/tidy3d/issues/1151#issuecomment-1734137975). Alec, could you take a quick look to see if it can simplify the kappa calculation in your case? No need to look into it if it doesn't work in a straightforward way.

tomflexcompute (approved):
The integration works great! I think in this implementation the following two lines aren't needed anymore? 

x_coords = sim_linear_data.monitor_data['field cavity'].Ez.x
y_coords = sim_linear_data.monitor_data['field cavity'].Ez.y

PR #267, opened 2023-09-08
adjoint example WDM

tylerflex:
Adds notebook showing the inverse design of a wavelength division multiplexer.


tomflexcompute (commented):
Thanks @tylerflex for another great adjoint example. See the comments below:

1. For the H1 header I think we have consistently only capitalized the first word. 

2. The link for "intro to inverse design tutorials" doesn't seem to work. Could you double check?

3. The comment for "lz = td.inf" is "# in 2D this is the". It seems incomplete? 

4. The BinaryProkector is imported but doesn't seem to be used. The projection is still a hand-coded function. 

5. Might want to make a comment about "params", which is a 1D array of length nx*ny that stores values between -0.5 and 0.5?

6. The header "Define base Simulation" -> "Define Base Simulation". "Define Design region" -> "Define Design Region".

tylerflex:
Thanks tom, I made all edits (nice catches..) and also fixed that broken link in the waveguide bend notebook. seems it was pointing to an earlier link when inv des 101 and fdtd 101 were in the same page.

tomflexcompute (approved):
Great. Thanks!

PR #268, opened 2023-09-08
replace filter feature_size with proper radius

tylerflex:
This PR just makes the notebooks compatible with the changes in the front end with the `ConicFilter`, which accepts `radius` instead of `feature_size`. note that the `radius` needs to be `sqrt(3)` times larger than the previous `feature_size` to get the same effect due to how we define things (radius=radius of conic filter, feature_size=heuristic for how large the features actually turn out).

@momchil-flex Ideally we should merge this before doing notebook runs for 2.4

### DIFF
--- a/docs/source/notebooks/AdjointPlugin3InverseDesign.ipynb
+++ b/docs/source/notebooks/AdjointPlugin3InverseDesign.ipynb
@@ -245,7 +245,7 @@
     "\n",
     "    values = jnp.array(eps_boxes).reshape((nx, ny, 1, 1))\n",
     "    \n",
-    "    conic_filter = ConicFilter(feature_size=0.15, design_region_dl=size_box_x)\n",
+    "    conic_filter = ConicFilter(radius=0.26, design_region_dl=size_box_x)\n",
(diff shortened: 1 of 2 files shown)


momchil-flex:
Looks good, but should we explain somewhere this relationship between radius and expected feature size?

tylerflex:
I added a few sentences to the definition of `radius`.
https://github.com/flexcompute/tidy3d/pull/1149

momchil-flex:
Right now it says "Note: the corresponding feature size expressed in the device is typically sqrt(3) times the radius (about twice the radius)." but isn't it the other way round, it's sqrt(3) smaller?

tylerflex:
you're right, I'll change that. good catch.

tylerflex:
changed to
```py
description="Radius of the filter to convolve with supplied spatial data. "
        "Note: the corresponding feature size expressed in the device is typically "
        "sqrt(3) times smaller than the radius. For best results, it is recommended to make your "
        "radius about twice as large as the desired feature size.",
```

PR #271, opened 2023-09-11
Minor notebook fixes

tomflexcompute:
- Fix header capitalization in adjoint waveguide bend notebook.
- Added SEO metadata in adjoint WDM notebook.
- Added SEO metadata in metadata.txt
- Fix header subscript in MoS2 notebook. Fixed two links in the notebook.
- Modified YJunction notebook to improve its workflow.

I changed the MoS2 in the notebook title from MoS$_2$ to MoS<sub>2</sub>. The two incorrectly rendered links in the notebook are related to using subscript and math character. Now they have changed to HTML formatting. I tried to build it locally but it didn't work due to some pydantic errors but hope it works in readthedocs build.


momchil-flex:
Was this at some point correct on 2.4 branch but then lost?

tomflexcompute:
Yeah not sure. The version in this branch is probably correct but we should double check.

tomflexcompute:
I took a closer look at the updated material_library.rst in this PR and it does seem correct.

PR #274, opened 2023-09-12
Added mm scale LNO coupler notebook

tomflexcompute:
This notebook is to be used in an LFW technical article that we are writing. The idea is to demonstrate mm scale photonic device simulation.


e-g-melo (approved):
Thanks Tom! Demonstrations in LNO are really necessary and the notebook also shows how to estimate task cost, which is indeed helpful for this large simulations.

I have only some minor comments:
1 - I suggest to include some information in the figure, like x,y,z directions, waveguides 1,2,3, optical axis, no, and ne directions, etc.
2 - I am not sure but it seems that Fig 1(c) of the paper shows no=2.138 and ne=2.211 (image quality is not that good).
3 - Exclude an empty cell in the end of the notebook.

e-g-melo:
Another suggestion: In the paper we find LN (Lithium Niobate) or LNOI (Lithium Niobate On Insulator). I suggest to use one of them in replacement of LNO.

tomflexcompute:
Thanks @e-g-melo for the helpful comments. Modifications have been added to the notebook. 

I think the paper marks n_e and n_o incorrectly. I checked multiple other sources and can confirm that the way in the notebook should be the correct one.

tomflexcompute:
Sorry @weiliangjin2021 I'll merge this in right now since the publisher needs a link urgently. Feel free to take a look whenever you have time and provide comments. I can modify it further after it's merged.

PR #275, opened 2023-09-13
Added PSR based on 90 degree bend notebook

tomflexcompute:
This can be potentially used in Momchil's meeting to work as a simple demo for tidy3d. It demonstrates 

1. How to use (remote) mode solver.
2. How to use gdstk to define bendy structrues.
3. Using ModeSource and ModeMonitor, a common approach in PIC simulations. 
4. How to plot 2D and 3D.


e-g-melo (approved):
Pretty good! 

My only comment is that the plot_3d() function on cell [8] did not show any image.

tomflexcompute:
Thanks @e-g-melo for reviewing so many notebooks recently :blush:

Yeah plot_3d() result doesn't show in the notebook but only shows when the docs are built.

momchil-flex:
This is great! I bet it will be a nice surprise for the meeting.

One comment: I feel like it's nicer to plot the modes as we do in other notebooks, with Ex and Ez profile for each mode. When plotting the intensity, it is really not clear what is TE and what is TM.

tomflexcompute:
> This is great! I bet it will be a nice surprise for the meeting.
> 
> One comment: I feel like it's nicer to plot the modes as we do in other notebooks, with Ex and Ez profile for each mode. When plotting the intensity, it is really not clear what is TE and what is TM.

Thanks. I've added field component plots. I think in general field norm or intensity plots are even easier to recognize. For standard waveguides at least, TE and TM modes are quite distinguishable just by looking at the field norm. In other mode solvers like lumerical or comsol, the default plot is also field norm so I think that would be good to have. It's also great to demonstrate how to plot field components as some users might want to do it.

momchil-flex (approved):
Great, thanks!

PR #276, opened 2023-09-13
Inclusion of y-branch sevel set optimization notebook.

e-g-melo:
This example demonstrates how to implement a simple parameterized level set optimization using jax. It seems a good option to further optimize an existing structure. 
Notes: The grid spacing on JaxCustomMedium is smaller than the simulation grid spacing over the design region. That worked fine, and allowed me to obtain a higher resolution level set surface, which was important to reduce the impacts of the grayscale generated by the hyperbolic tangent step function.


tomflexcompute:
Only took a quick look but this looks very nice! I'll take a closer look later today or early tomorrow. The final design is very interesting looking. The junction part is almost disconnected from the output waveguides yet it works very well.

tomflexcompute:
Thanks @e-g-melo for this great demo. My expertise in topological optimization is rather limited but clearly it works very well in this case. How long does it take to run the notebook btw? Is this method inherently difficult to work with symmetry?

A few typos that I found:

1. Comment on cell[7], "more than one evaluations" -> "more than one evaluation".
2. Above [13],  "The elements that does not change" -> "The elements that do not change".
3. Above [14], "is build" -> "is built".
4. Above [18], "are the coupling efficiencies" -> "is the coupling efficiencies". "imbalance" -> "imbalance".

Also we can use a thumbnail image like this. Currently all adjoint examples do not show a thumbnail image in the Learning Center but I suspect we will change that in the future. Having a thumbnail image always helps when sharing the example in social media anyway so let's put one whenever possible.
ybranch_level_set

tomflexcompute (approved):

e-g-melo:
Hi Tom! Thank you for the comments! The typos were corrected.

To implement symmetry basically we need to make some changes in `update_design` using the same approach employed on GC adjoint notebook. I put symmetry only about the z-axis to enable different power fractions on the upper and lower waveguides, so someone can quickly convert this y-branch to a MMI power splitter. The total optimization took about 1 hour here.

The thumbnail image is great! Thank you for that. I wonder if we could smooth the sidewalls. I tried that using the code below, but it introduced some deviations on layout.

`import gdstk
from shapely import Polygon

# Get the contours of the permittivity distribution.
sim_domain = td.Box(center=(0, 0, 0), size=(size_x, size_y, 0))
eps_final = sim_final.epsilon(box=sim_domain, coord_key="centers")
eps_contours = gdstk.contour(np.squeeze(eps_final).real.T, 
                            level=(eps_max + eps_min)/2, 
                            length_scale=grid_size,
                            precision=grid_size/100)

# Use shapely to smooth the sidewalls.
poly_0 = Polygon(eps_contours[0].points)
poly_1 = poly_0.simplify(grid_size/3, preserve_topology=True)

# Export to GDS.
lib = gdstk.Library()
cell = lib.new_cell("Y-BRANCH")
cell.add(gdstk.Polygon(poly_1.exterior.coords))
lib.write_gds("./misc/ybranch.gds")`

tylerflex:
> @tylerflex, can we have a jax version of the function gdstk.contour. If that is possible, I think we could use the same minimum curve radius you wrote for shape optimization to impose fabrication constraints on level set.

I would love that, but no idea how we could implement it. Maybe @lucas-flexcompute would know. does `gdstk.contour` comprise of differentiable functions that could replace `numpy` with `jax.numpy`?  or would it be more involved? 

@e-g-melo do we have any degrees of freedom in the level set function that could impose some fabrication constraints already? For example, restricting the local curvature of that function? Or choosing a function that is more slowly varying in space?

tylerflex:
Thanks @e-g-melo I'm really excited to have this demo! A lot of people ask about this method and this seems to work quite well!

I had a few comments

* I’ve been importing the adjoint plugin as `import tidy3d.plugins.adjoint as tda` and it makes life a lot easier, eg `tda.JaxSimulation`. I would recommend this for notebooks moving forward.
* Is there a reason for using scipy's LBFGS here? I’ve been having better luck with `optax` for simplicity and more access to the optimizer state within runs for checkpointing and debugging.
* before [14] “And then the JaxSimulation is (build -> built) using the design parameters.”
* I think a bit more discussion or explanation of the level set technique might be useful in this notebook. So it serves as a bit of a tutorial for people who are less familiar with it. So as far as I understand it, the implementation here essentially starts with a function f(x, y), which is parameterized using a gaussian basis. We take f > 0 to be defined as material and f < 0 as background. Our free parameters are essentially the amplitudes of each of the gaussian basis functions? Are there other choices of basis functions that people typically use? is there a good reference to cite? What are the typical ways of evaluating the f=0 level set in a differentiable way? Maybe I am a bit confused on some of these points so it could be useful to have some a little more explanation in the notebook.
* What explains the slight asymmetry in the final structure? Should we try with symmetry on? Or does that complicate things too much?

* Did we ever resolve the issue with these few lines of code
```py
      # Calculates the permittivities from the level set surface.
    sharpness = 10
    eps_phi = 0.5 * (jnp.tanh(sharpness * phi) + 1.0001)
    eps = eps_min + (eps_max - eps_min) * eps_phi
```
The choice of `tanh()` here was just my suggestion to make this a "smooth" process that is differentiable by jax, but it likely is non-stanrdard. What is the typical approach to defining the permittivity from the level set function in practice? And how does one differentiate through this step?  For example, one could imagine using `gdstk.contour` to get a polygon and then using adjoint boundary gradient formalism to compute the gradients, but how does one typically define the polygon from the raw phi data? I guess I’m a little confused about what is the standard approach to this step and want to make sure we at least understand what is the proper way first and then also add a bit of explanation so the user also knows what’s going on here. I think what the tanh approach does is creates a thin "grayscale" pixel layer between the material and background, which can be differentiated through the `JaxCustomMedium`, but a more standard approach might involve defining a sharp boundary and differentiating through the moving boundaries formalism so I want to be sure I understand.

Thanks!

lucas-flexcompute:
> I would love that, but no idea how we could implement it. Maybe @lucas-flexcompute would know. does `gdstk.contour` comprise of differentiable functions that could replace `numpy` with `jax.numpy`? or would it be more involved?

Unfortunately, no. It implements the [marching squares](https://en.wikipedia.org/wiki/Marching_squares) algorithm with some modifications to take care of edge contours. I can't see it being differentiable.

tylerflex:
> Unfortunately, no. It implements the [marching squares](https://en.wikipedia.org/wiki/Marching_squares) algorithm with some modifications to take care of edge contours. I can't see it being differentiable.

Yea that sounds hard. The only way I could see this working is if we take the `td.PolySlab` constructed by `gdstk.contour` and connect the derivative of the polyslab vertices to the change in the underlying parameters (maybe using a linear fit?). It sounds challenging but it would be good to know what is the standard practice here for computing this gradient to understand that.

lucas-flexcompute:
I think this case is usually only dealt with using a level-set function.

tylerflex:
What's the status of this PR? Maybe @e-g-melo you can add a short discussion about choice of tanh() and we can try to merge this soon?

e-g-melo:
Sure, doing that today..

e-g-melo:
Hi @tylerflex and @tomflexcompute . Thanks for your comments!

I've updated the notebook to correct the typos and included an image of the final structure following the Tom's suggestion. 

Regarding the Tyler comments:

- I’ve imported the adjoint plugin as import tidy3d.plugins.adjoint as tda.

- There was not a specific reason for using LBFGS, so I updated the notebook including the Adam optax optimizer. Its convergence is a bit slower than LBFGS, so I increased the number of iterations a bit. However, the final structure is better now, it has smoother lines and is more symmetric than before.

- I included more details and references about the level set implementation. Actually, a parameterized level set implementation using `jax` is more simple, because we don't need to track and calculate the gradients of the moving boundaries by ourselves, which is the standard approach. So, it resumes to create the level set surface given the design parameters (which are the control knots), and then map the permittivity to the zero contour. There are other alternatives in addition to the Gaussian RBF used to fit the design parameter and create the level set surface. I included some references to them.

- To include symmetry we need to modify the notebook in the same way as the grating coupler example. I've not included that to make things simple and to allow the user change the objective function if splitting rations other than 50/50 are desired.

- I've tested a sigmoid function in addition to the tanh one, but `jax` seems to produce bit more stable results using the tanh when using higher slopes. There are some level set implementations which uses a hard step function, and other ones using this type of smooth functions. I included a short discussion about the tanh() function and a reference to a paper which states that the differences between the tanh, sigmoid, and arctan functions reduce with grid size.

- In our implementation we have some control over the minimum feature size by changing the distance between the control knots and the width of the Gaussian functions. As not perfect, it goes in the same direction of this paper (https://opg.optica.org/oe/fulltext.cfm?uri=oe-28-5-7060&id=427821), where higher order b-splines are used to control the minimum feature sizes. There are some papers proposing the introduction of fabrication constraints in level set optimizations using an approach similar to the density-based one (https://link.springer.com/article/10.1007/s00158-020-02527-1). And this one from Jelena group (https://www.nature.com/articles/s41598-019-45026-0) which present penalization functions to the minimum feature size and curvature. In a second stage we can try to test the first two and see how them perform.

tylerflex (approved):

PR #277, opened 2023-09-14
Low loss material in material_library

weiliangjin2021:
When using tidy3d from my weiliang/max_loss branch, the built doc now has 3 types of materials in the material library table: lossless, lossy, and low loss. As I look through the page, all Palik lossless materials and MgO are marked as low loss.


e-g-melo (approved):
Thank you Weiliang!

PR #280, opened 2023-09-15
Added PSO of a PBS notebook

tomflexcompute:
The particle swarm optimization demo was used for a customer demo and now can be published as a public example. Also addressing https://github.com/flexcompute/tidy3d/issues/1107. 

@e-g-melo might have a second PSO example on a different device, which can be published later when ready.


lucas-flexcompute (approved):
Looks great. Just a few minor comments:
- In line 5 in the introduction, "We" shouldn't be capitalized.
- In input 4, we use capitalized variable names. That's not a common practice in python, but I can see it being useful to match the original paper if that's the case. Do we follow this convention in other notebooks as well?
- Instead of using `PolySlab.from_gds`, I think we should start using `Geometry.from_gds`, as it fully supports self-intersections, holes, etc.
- It might make sense to define the whole taper + S-bend + output waveguide as a single `gdstk.RobustPath`
- In the closing remarks there's a missing link for the publication.

momchil-flex:
Awesome. Why is the result only a moderate improvement compared to the paper? What would the user need to do to get better results - more particles, more line segments? (worth commenting on)

One annoyance that emerged when using the adjoint plugin was that if you get disconnected in the middle of the optimization, you can't continue it. This is why when we started using the optax plugin, we also showed how one can save the optimization at every step (it has an `optimizer.step` method that can be used instead of `optimizer.optimize`). However, I don't see a way to do this with pyswarms unfortunately...

Could you also just black the notebook? There are a few lines of code that are very long and go outside of the code box when viewed on github. May have similar issue in our docs.

tomflexcompute:
Thanks @lucas-flexcompute for the great suggestions. Modifications have been made. 

Indeed I usually try to match the variable naming convention of the paper so it's a bit easier for readers to identify them.

tomflexcompute:
Thanks @momchil-flex . In the paper they used a "two-stage" optimization where the first stage optimizes W1 to W7 with one FOM and then W8-W10 are optimized with a different FOM to minimize crosstalk, if I understand correctly. They also used a much larger number of particles and iterations I believe. 5 particles and 40 iterations are definitely on the small side for PSO. We have a comment at the intro and closing remark to remind the user that a more elaborate PSO scheme should be used to achieve better performance. Now I also just added a comment on using more particles and iterations. 

Actually internet disconnection did happen once during a run. I saw the message saying that it's trying to reconnect in 180 seconds. Eventually it did reconnect and the run continued. I think it's pretty good. If the internet is disconnected for over three minutes I guess it is a bit hard to recover. PySwarms doesn't seem to provide a way to store intermediate particle positions as far as I see. 

And yes I usually black them right before merging. It's reformatted now.

e-g-melo (approved):
Thank you Tom for this notebook! It is very didactic and informative. I have no comments here.

PR #283, opened 2023-09-18
adjoint with better filter size constraints

tylerflex:
Modifies the mode converter and WDM notebooks:
* two rounds of filter -> projection
* clean up some of the flow in the mode converter notebook (make basic sim, run mode solver, add mode mnt, src).
* penalty to objective if erosion -> dilation of device is not equal to dilation -> erosion, which is a simple implementation of minimum length scale.

I'm re-running these notebooks with larger step_size and smaller features (to get better results) so this commit is mostly a placeholder to save results, shown below.

image
image


tomflexcompute:
It looks like the initial condition is using a random distribution without a fixed seed. Would rerunning the notebook leads to a very different result very time?

tylerflex:
That's correct (and was true before), but I can add a seed. I'm planning to re-run these this morning / afternoon so I'll add that.

tylerflex:
I ran this again and the results are better:
* mode converter was helped by ramping up beta and the penalty for feature size.
* WDM actually didn't work at all with that change, but benefited from a very small step size.

image


image

PR #285, opened 2023-09-19
Inclusion of bullseye cavity PSO optimization.

e-g-melo:
PSO optimization of a Bullseye cavity. This notebook has two purposes: (i) show another example on how to set up and run PSO; (ii) show an example on how to use Tidy3D to build simulations related to light extraction from quantum emitters embedded/over cavities, which could find potential applications in single-photon sources, LEDs, etc. I've tried to be as general as possible.

The notebook is also related to Particle swarm optimization #1107.


tomflexcompute:
This is a very nice example @e-g-melo ! Light extractor is a great addition to our example library and Purcell factor calculation has also been asked by a few users from time to time as well. A few minor comments for your consideration:

1. "inspired in" -> "inspired by" ?
2. "centre" -> "center"
3. "aiming practical applications" -> "aiming for practical applications" ?
4. "increase the number of iterations" -> "increasing the number of iterations"
5.  It seems like the objective function reaches a plateau after the 12th iteration? This is likely due to the number of particles is too small. Should we just run like 15 iterations to save time or use a different random seed to see if it improves?

momchil-flex:
Yeah, agreed with Tom, this is great and it is also high time we had a Purcell factor demonstration!

> It seems like the objective function reaches a plateau after the 12th iteration? This is likely due to the number of particles is too small. Should we just run like 15 iterations to save time or use a different random seed to see if it improves?

I agree with this point; as is, the optimization could be cut to 15 iterations. This is probably fine, and you can just comment that e.g. using a larger number of particles may help.

It should actually be possible to get the Purcell factor without a normalizing simulation. Our dipoles are normalized such that in a homogeneous material they would inject power as given by the formula [here](https://docs.flexcompute.com/projects/tidy3d/en/latest/faq.html#source-power-normalization). There could be a small difference due to the numerical grid because this formula is only valid in the case of infinite resolution. But I would be curious if you can use this directly and get the same result. Maybe it's a good opportunity to show both ways and comment that the normalizing one requires an extra simulation but is slightly more accurate?

e-g-melo:
Hi @momchil-flex! Thanks for your comments!

I've reduced the number of iterations to 15, as you and @tomflexcompute suggested.

When introducing the analytical dipole power calculation I found a divergence between the dipole power obtained with and without symmetries. Is that expected? See below:
Picture1

It seems the dipole power increased by ~4 when we have 2-fold symmetry and by ~16 when using 4-fold symmetry.

e-g-melo:
4-fold symmetry
image

momchil-flex:
Ah. Yes, this is expected, but it does complicate things. The motivation for this was continuity. Essentially, a dipole sitting on a symmetry plane is equivalent to *two* dipoles (and when it sits on two symmetry planes, it's correspondingly equivalent to 4 dipoles). Consider the case in which the dipole is slightly off the symmetry plane, at some x > 0. Then, depending on whether you put symmetry or not in the simulation you would actually get different results: the non-symmetry one has a single dipole at x; however, the symmetry one effectively has a dipole at x and an "image charge" at -x. As you move x towards 0, the dipole and its image get close together. In the limit of x = 0, they just lie on top of each other. We could have made a choice to ignore this when the dipole is sitting exactly on the symmetry plane and treat it as a single one, but then that leads to other potential issues, like a discontinuity in the radiated power when you move the dipole towards the symmetry plane.

I don't know if this was a correct choice to make... but it's how it is right now. So maybe this is also a good opportunity to explain that? Basically the dipole power should be scaled by something like `2 ** np.sum(np.array(sim.symmetry) != 0)`.

tomflexcompute:
> Ah. Yes, this is expected, but it does complicate things. The motivation for this was continuity. Essentially, a dipole sitting on a symmetry plane is equivalent to _two_ dipoles (and when it sits on two symmetry planes, it's correspondingly equivalent to 4 dipoles). Consider the case in which the dipole is slightly off the symmetry plane, at some x > 0. Then, depending on whether you put symmetry or not in the simulation you would actually get different results: the non-symmetry one has a single dipole at x; however, the symmetry one effectively has a dipole at x and an "image charge" at -x. As you move x towards 0, the dipole and its image get close together. In the limit of x = 0, they just lie on top of each other. We could have made a choice to ignore this when the dipole is sitting exactly on the symmetry plane and treat it as a single one, but then that leads to other potential issues, like a discontinuity in the radiated power when you move the dipole towards the symmetry plane.
> 
> I don't know if this was a correct choice to make... but it's how it is right now. So maybe this is also a good opportunity to explain that? Basically the dipole power should be scaled by something like `2 ** np.sum(np.array(sim.symmetry) != 0)`.

That's interesting to know. When the dipole is not sitting on the symmetry plane, inevitably it's equivalent to two dipoles. But when it is sitting on the symmetry plane, I think it's considered as one dipole in other software. So this scenario where you continuously move a dipole from outside of the symmetry plane to on the symmetry will lead to a discontinuity. I personally would prefer this as it's more physical to me. Since it's not the case right now, I agree that we should add a comment about it here to clarify and use it to educate users.

e-g-melo:
Thanks @momchil-flex and @tomflexcompute! I've introduced the changes on the power calculation and some comments about the dipole image charge.

tomflexcompute (approved):
Thanks @e-g-melo for the update. Looks really good. Just wondering after reducing the iteration to 15, does it still cost more than 15 credits?

e-g-melo:
My bad! Actually it costs less then 2 FlexCredits now. That cost was still related to the first initial version where I was using a higher-Q cavity.

momchil-flex:
> > Ah. Yes, this is expected, but it does complicate things. The motivation for this was continuity. Essentially, a dipole sitting on a symmetry plane is equivalent to _two_ dipoles (and when it sits on two symmetry planes, it's correspondingly equivalent to 4 dipoles). Consider the case in which the dipole is slightly off the symmetry plane, at some x > 0. Then, depending on whether you put symmetry or not in the simulation you would actually get different results: the non-symmetry one has a single dipole at x; however, the symmetry one effectively has a dipole at x and an "image charge" at -x. As you move x towards 0, the dipole and its image get close together. In the limit of x = 0, they just lie on top of each other. We could have made a choice to ignore this when the dipole is sitting exactly on the symmetry plane and treat it as a single one, but then that leads to other potential issues, like a discontinuity in the radiated power when you move the dipole towards the symmetry plane.
> > I don't know if this was a correct choice to make... but it's how it is right now. So maybe this is also a good opportunity to explain that? Basically the dipole power should be scaled by something like `2 ** np.sum(np.array(sim.symmetry) != 0)`.
> 
> That's interesting to know. When the dipole is not sitting on the symmetry plane, inevitably it's equivalent to two dipoles. But when it is sitting on the symmetry plane, I think it's considered as one dipole in other software. So this scenario where you continuously move a dipole from outside of the symmetry plane to on the symmetry will lead to a discontinuity. I personally would prefer this as it's more physical to me. Since it's not the case right now, I agree that we should add a comment about it here to clarify and use it to educate users.

In hindsight I probably would have taken the other approach and break continuity in favor of obtaining the same results with or without symmetry. Maybe we can change this in 3.0.

PR #294, opened 2023-09-29
Added MIM waveguide sensor notebook

tomflexcompute:
Added MIM waveguide CO2 sensor example recommended by an user.


tomflexcompute:
Thanks @alec-flexcompute for looking into it! These are great points. Will implement the changes once Emerson's comments come in.

e-g-melo (approved):
Hi Tom, thanks for another pretty good example! Congrats.

I have just a few comments on it:
1) Your figure is very nice!! I wonder how you have overlapped the field image over it. Was it made using Blender?
2) Would it be a good opportunity to use SiO2 Palik? Will it change the result?
3) I suggest you to include tabulation in `sim` definition (cell 7).
4) It seems the resonance spectra are a bit different from the paper ones (Fig 6(c)). Do you know why?
5) Do you know why we are getting "Output()" after task submission?

tomflexcompute:
Thanks @e-g-melo for the review and comments. 

- The 3D rendering is done in Blender but the red and blue field pattern is actually made in powerpoint. I just made some semi-transparent blocks with a gradient color. Surely there are better ways to make it. 

- Yes it's a great idea to use Palik. It doesn't change the result by much. 

- I'm a bit confused by what you mean by "tabulation in `sim`"? Could you clarify? 

- It's a great observation. The paper is actually using 2D simulation as far as I can tell. I would say our 3D results are more reliable. Also the refractive indices of the metal is probably slightly different as they didn't disclose the details either. 

- Yeah sometimes these "output()" show up for some reason. Locally I don't see them. I think it's related to how github converts notebook file to web page.

e-g-melo:
Sorry.. I mean include indentation.

PR #296, opened 2023-09-29
WDM and multi-objective using broadband adjoint

tylerflex:
Notebook changes corresponding to https://github.com/flexcompute/tidy3d/pull/1192

1. The multi-objective (adjoint plugin 4) was changed to a backend test since we no longer need batching over frequency. This PR changes it to instead do the batching over the box size in y. Eventually we probably want to do a more sophisticated example involving dilation and erosion in the objective fn.
2. The WDM optimization was changed to use the broadband adjoint feature instead of batching over frequency.


weiliangjin2021 (approved):
All look good. Two minor questions in WDM notebook:

- In the function `pre_process`, why is filter_project applied twice?
- In `make_eps`, why is the base permittivity 1.0001?

tomflexcompute:
Thanks for the update. 

Here should we just remove the commented codes? 
image
Also in the comment it's mentioned that the cross-transmission is subtracted. However, in the new implementation, it doesn't seem to be the case?

tylerflex:
@weiliangjin2021 

> * In the function `pre_process`, why is filter_project applied twice?

I found that doing the filter project twice with smaller beta (projection strength) values tends to produce better results. It is a trick that is mentioned in some Sigmund papers.

> * In `make_eps`, why is the base permittivity 1.0001?

This is because if I set it to 1.0 exactly, there can sometimes be numerical issues that set the permittivity slightly less than 1. 

@tomflexcompute thanks for pointing this out, I actually forgot to push the latest version which cleans these things up.

weiliangjin2021:
> I found that doing the filter project twice with smaller beta (projection strength) values tends to produce better results. It is a trick that is mentioned in some Sigmund papers.

I see, nice trick. Would it help to write it in the fashion of loop to highlight it's done deliberately, and also that one can easily apply this more than twice if needed in some optimization?

> This is because if I set it to 1.0 exactly, there can sometimes be numerical issues that set the permittivity slightly less than 1.

Will apply `max(1, eps)` by the end help?

tylerflex:
> Will apply max(1, eps) by the end help?

that would work, but I feel might introduce a bit of discontinuity in the values. Let me look into this a bit more, I think maybe the 1.001 is not really needed if the parameters themselves are constrained between 0 and 1.

> Would it help to write it in the fashion of loop to highlight it's done deliberately, and also that one can easily apply this more than twice if needed in some optimization?

I've only seen it done twice maximum, but I get I could do something like
```py
num_filters = 2

val = params
for _ in range(num_filters):
    val = filter_project(val)

```

if you think that makes it clearer

tomflexcompute (approved):
Thanks. A few users seem to have trouble directly converting our 2D adjoint examples to 3D for real device design. What's the inherent issue here?

tylerflex:
The inherent issue I think is that 2D devices seem to offer more control over the device functionality in a finite design area. When the devices become 3D, they generally need to become larger in area to accomplish the same performance. This is something we observed in our other WDM case. So I think this can be the general recommendation.

momchil-flex (approved):
> The multi-objective (adjoint plugin 4) was changed to a backend test

Where? Or is this yet to come?

Otherwise, I like the modification to the notebook to discuss handling fabrication tolerance in this way!

Regarding this warning:

```
WARNING: 2 unique frequencies detected in the output monitors with a 
           minimum spacing of 3.720e+13 (Hz). Setting the 'fwidth' of the       
           adjoint sources to 0.1 times this value = 3.720e+12 (Hz) to avoid    
           spectral overlap, which will introduce significant error in the      
           gradients if not avoided. A small 'fwidth' will automatically set a  
           correspondingly long 'run_time' in the adjoint simulation, which may 
           cause it to run significantly longer than  the forward simulation in 
           some cases. If your output frequencies are tightly packed, it would  
           be preferrable to instead run one simulation per frequency using     
           'tidy3d.plugins.adjoint.web.run_async', potentially, to run these    
           simulations in parallel. 
```

I wonder if we can rephrase a bit to make more helpful (and concise in places). Something like:

```
WARNING: 2 unique frequencies detected in the output monitors with a 
           minimum spacing of 3.720e+13 (Hz). Setting the 'fwidth' of the       
           adjoint sources to 0.1 times this value = 3.720e+12 (Hz) to avoid    
           spectral overlap. The corresponding 'run_time' in the adjoint simulation,
           is {...} compared to {...} in the forward simulation. If the adjoint 'run_time'
           is large due to small frequency spacing, it could be better to instead run one 
           simulation per frequency, which can be done in parallel using 
           'tidy3d.plugins.adjoint.web.run_async'.
```

tylerflex:
>> The multi-objective (adjoint plugin 4) was changed to a backend test
> Where? Or is this yet to come?

It's currently in the front end PR as an underscore file in `tests/test_plugins/` but I'll move it to a core PR after this.

I think the modified warning message is an improvement, I'll implement that in the front end PR and fix this notebook up, thanks.

tylerflex:
Actually, in the interest of not running the entire notebook over again, maybe we can just leave the current output? 

image

Unless you think it's worth it to re-run, I might just leave it for now until the next time this is run.

momchil-flex:
> Actually, in the interest of not running the entire notebook over again, maybe we can just leave the current output?
> 
> image
> Unless you think it's worth it to re-run, I might just leave it for now until the next time this is run.

Not super important, but why not just rerun up to that message (but not the optimization)?

But also actually you should probably rerun it after the solver is rebuilt to use server instead of local run right?

tylerflex:
| Not super important, but why not just rerun up to that message (but not the optimization)?

I think if I restart the notebook, the plots and results will not be displayed. (at least I think the cell numbers will be out of order).

| But also actually you should probably rerun it after the solver is rebuilt to use server instead of local run right?

Yea I'll do that, could you or @dbochkov-flexcompute rebuild the solver with the current state of the front end pre/2.5 (maybe with a new mapping for 2.5.0rc2) so I can give this a try? thanks

tylerflex:
Alright, I just changed the local runs to server side runs, tested these notebooks over the webapi, tested that the outputs were the same. Then I ran the WDM notebook only up to where the warning is created and the notebook is fine (subsequent output and line numbers are still good). So this should be ready to go.

PR #297, opened 2023-10-01
Added trimesh demo

tomflexcompute:
Added the trimesh demo to show how to define complex geometries and apply transformations that are hard to achieve natively in tidy3d. 

The 3D view of the mesh is not displayed on Github but I think it should display correctly once the docs are built.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -47,6 +47,7 @@ Structures
     notebooks/SelfIntersectingPolyslab
     notebooks/PICComponents
     notebooks/PhotonicCrystalsComponents
+    notebooks/CreatingGeometryUsingTrimesh
 
 Grid Specification
 ~~~~~~~~~~~~~~~~~~
--- a/docs/source/notebooks/metadata/metadata.txt
+++ b/docs/source/notebooks/metadata/metadata.txt
@@ -149,6 +149,12 @@
     "keywords": "silicon photonics, PIC, integrated photonics, bullseye cavity, quantum emitter, particle swarm optimization, PSO, Tidy3D, FDTD",
     "feature_image": "./img/bullseye_psa_thumb.png"
   },  
+  "CreatingGeometryUsingTrimesh.ipynb": {
+    "title": "Defining complex geometries using trimesh | Flexcompute",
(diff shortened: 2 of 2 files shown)


shashwat-sh (commented):
Looks great! I think this will be very useful. Some minor suggestions below (optional, I don't think they are crucial):
- In cell 7 where the box is created from a set of vertices: when defining the faces, can we clarify whether the vertices need to be set in any particular order, e.g. to make sure the normal vector always points in or out? Or if there is no such requirement, can we mention that?
- For the torus and ellipsoid, it seems these cells are demonstrating features of Trimesh more so than Tidy3D. It might be nice, at the end each of these examples, to have the Tidy3D `from_trimesh` function call (as with previous examples) just to bring the focus back to Tidy3D and conclude the section by saying "after creating this complex shape, this is how we can import it into Tidy3D". One idea could be to import both the torus and ellipsoid (cells 9 and 11) into the same Tidy3D sim, and then use our plotting to plot both in the same domain, to reinforce the point that one can create any number of these complex shapes and put them all together easily.
- On a related note, I wonder if we should add a link to Trimesh in case users want to access their documention for how to make more complicated shapes?

tomflexcompute:
Thanks @shashwat-sh for the great suggestions. 

- I don't think the order matters here when generating the mesh. However I'm not sure if the normal vector direction will cause any issues in tidy3d? 
- That's a good idea. I put the torus and ellipsoid into a simulation and plotted it. I noticed that plot_3d() shows nothing while 2D plot works fine. Is it expected? @caseyflex 
- Links to trimesh docs have been added to everything trimesh mention.

shashwat-sh:
Thanks, looks great to me!

caseyflex:
The order of the vertices does matter. For example, in your box, changing the front face from `[0,1,2]` to `[1,0,2]` messes up the box. Then, tidy3d would give a warning about inconsistent face orientations. So it might be nice to comment about this, yeah

tomflexcompute:
Thanks @caseyflex . Indeed that's the case. 

Regarding the 3D plot, it seems like if I import a stl file, it will show correctly, but a geometry defined from a trimesh object doesn't. Don't they use the same plotting mechanism?

PR #298, opened 2023-10-02
Adding mode info dataframe output to some notebooks

momchil-flex:
I am not sure if there's a way to avoid the unnecessary notebook diffs.. I need to rerun cells up to the mode computation, and that creates some diffs. Anyway, the things that I actually changed should be only around places where there's a `to_dataframe()` call - you can focus on those.

Any other notebooks you think would be useful to add this to?


tomflexcompute (approved):
Very nice! Thanks for this addition. I think showing the dataframe output in these notebooks is sufficient.

PR #299, opened 2023-10-03
Inclusion of Symmetry tutorial.

e-g-melo:
This is the symmetry tutorial. I've included two examples and discussed how to identify PEC and PMC-like symmetry planes.


tomflexcompute (approved):
Thank you @e-g-melo for this great tutorial that will solve many users' confusion. It also makes great learning material for new users. 

I noticed that the featured_image metadata is empty while the metadata.txt has an image path in it. Usually I just copy one to the other so they are totally identical. 
image

In the waveguide example, can we add a comment to remind users that symmetry is only applicable when the sidewall angle is zero and the top cladding is the same material as the substrate? Otherwise only the symmetry y direction exists.

e-g-melo:
Thanks for reviewing it, Tom! I've included that observation about waveguide sidewall angle and the feature_image info.

PR #301, opened 2023-10-05
basic heat solver examples

dbochkov-flexcompute:


e-g-melo (approved):
Hi Daniil! Thanks for these notebooks! Congrats, this is an exciting new feature!

Overall, the notebooks are very nice. I have only some comments on them:
1) HeatSolver.ipynb
- We use a capital letter only at the first word of the notebook title.
- Headers H2 should have all the words with an initial capital letter. That would be needed to have a uniform format throughout the notebooks.
- Maybe change 'api' to "API"?
- It would be nice to have a comment about physical units when you define conductivities, temperatures, sources, etc.
- Is there something missing in this sentence, "First way if to use function web.run()"?
- I suggest plotting the numerical solution on top of the analytical one in the graph or any other way to make it easier to see both results. 
- It would be interesting to have some comments about the solution error graphs.

2) MetalHeaterPhaseShifter.ipynb
- Same comment about H1 and H2 headers.
- When you defined the material properties, you included their physical units, which is nice, but I suggest including short comments about the meaning of each parameter as well. 
- I wonder if the results would change if you don't ignore heat transfer between SiO2 and air.
- Organizing the notebook into subsections would be interesting so the readers can move into specific parts quickly. 

In addition, please take a look at the docs/source/notebooks/metadata/metadata.txt file to include SEO information there.

weiliangjin2021:
The notebooks look beautiful! Some comments for the 1st notebook (the 2nd one in next Monday):
- Below steady-state equation, define "s".
- Below "Defining simulation scene": add a schematics to illustrate the layers and definitions such as "r1", "conductivity1', etc.
- In Cell 2, add units to the parameters.
- If capacity has no effect, maybe not show it at all in SolidSpec or use some default values

weiliangjin2021:
The 2nd notebook looks good to me.

momchil-flex (approved):

dbochkov-flexcompute:
forgot to include some new files, force-pushed

weiliangjin2021 (approved):

PR #302, opened 2023-10-09
ran grating coupler with double filter

tylerflex:
Leads to a bit more fabricable structure, convergence in 75 iterations instead of 150. However, final result has -2.3 db of loss as opposed a little less than -2 db for the existing notebook.

Could merge this if we want, or I could run with a smaller feature size if desired. Just making the PR in case we want to include this technique.

Before:
image

After:
image


e-g-melo (approved):
Thanks for these improvements in GC notebook, Tyler! The final structure is very nice! 

In addition to reducing the feature size, maybe we can try a structure a bit larger using the same feature size as well. 

BTW, why is the number of points in a custom medium limited to 250,000?

tylerflex:
> Thanks for these improvements in GC notebook, Tyler! The final structure is very nice!
> 
> In addition to reducing the feature size, maybe we can try a structure a bit larger using the same feature size as well.

Do you think it's worth merging this now and then trying this next time we we-run notebooks? I might do a re-run after broadband adjoint is build on server side?

> BTW, why is the number of points in a custom medium limited to 250,000?

This is because the postprocessing time when the number of points is above this can become very slow (minutes I think). so we don't want to incur too many costs?

PR #305, opened 2023-10-12
Thermo-optic phase-shift in a ring resonator

dbochkov-flexcompute:
Feel free to give a feedback on the notebook's content and flow as well as the heat solver interface itself. Don't be shy to give suggestions that would require a good amount of content rework, because it's pretty basic at the moment.


tomflexcompute:
This is a wonderful example that is very typical. Results are great too. I think it serves as a popular example for users who want to do thermoptic simulations. 

A lot of newly introduced classes make the simulation setup quite a bit more complex but I suppose it's somewhat inevitable. I have the following suggestions after reading the notebook for the first time. It's quite new to me as well so will take a closer look later this week. 

- The heater is pretty close to the top cladding surface so it seems a bit unrealistic to me to set the temperature to 300 K on the top? If we include a larger air domain, will the temperature distribution be very different?

- The optical simulation inevitably will take a long run_time and thus large FlexCredits. Can you add a warning at the beginning of the notebook like we did in other large notebooks? Something like 
image

- The x and y sizes of the optical simulations have been reduced. In this case, I think the z size can be safely reduced as well. Probably 2 um in z would be sufficient I think. The Si substrate shouldn't have any impact on the result. 

- `num_modes=2` for the two ModeMonitors. Is there any particular reason why? I think we are only using the fundamental mode?

- One thing that is slightly confusing to me is that `heater_source = td.UniformHeatSource(structures=["heater_outer"], rate=volumetric_heat)`. Here the structure is set to the outer cylinder instead of the ring. Is it fine because the inner cylinder overrides the outer cylinder so it automatically knows the actual structure is a ring? Also one general question I have is how to pick a specific surface of a structure? For example, if I have a Cylinder inside a Box, MediumMediumInterface will be the outer surface of the cylinder. What if I want to apply certain boundary condition only on the top surface of the Cylinder?

lucas-flexcompute (commented):
That's a very cool example! It will be extremely useful as a starting point for many users.

A few things I'd change:

- It's probably more conventional to place either the center of the waveguides or their bottom surface at z=0. It makes plotting and geometry construction easier in general.

- For the rings, you can use our new hole support: `geometry = td.Cylinder(... outer radius ...) - td.Cylinder(... inner radius ...)`. This works even if there are other structures in the hole region.

- Is there any reason to create a separate `heater_gap` structure? Shouldn't `box_clad` be enough?

- The size of the optical simulation can probably be safely decreased to 1.5µm above and below the device layer. That should help with the simulation cost and time.

- Typo at "we enforcing approximately…"

- Is it possible to visualize the heat grid?

- The heat plot in cell 75 is not gray scale, as indicated in the comments and color bar. Same in cell 87.

- The creation of the perturbed optical sim with `exclude={"type"}` is very non-intuitive for users. Any chance to have a custom method like `optic_sim.with_scene(perturbed_scene)` or something along those lines?

- The gap between ring and waveguide is so small that it is exciting counter-propagating waves in the ring (see the standing-wave pattern in cell 97). It's also unreasonable from a fabrication point of view, so we should probably increase it to 100 nm, which is much more common. @tomflexcompute this is also a problem in the current ring resonator example, btw - I noticed that during ECOC and forgot to mention.

I also have another question: is it possible to define different boundary conditions to boundaries sharing the same material interfaces? For example, if the simulation did not include the bottom Si wafer, we'd have 2 SiO₂/air interfaces with different BCs. Is that possible to define using the `placement` argument in some other way?

dbochkov-flexcompute:
> The heater is pretty close to the top cladding surface so it seems a bit unrealistic to me to set the temperature to 300 K on the top? If we include a larger air domain, will the temperature distribution be very different?

Actually we don't set temperature of the cladding top surface. What we impose there is the so called convective cooling boundary conditions. It states that the heat flux through the surface is proportional to the difference between the surface's temperature and temperature of the surrounding fluid away from interface, that is, $q = h (T - T_{air})$. This can be derived from fluid dynamics using some approximations. So, we set the temperature of the surrounding fluid $T_{air}$, but the temperature of the surface is found as part of solving the heat equation. The heat transfer coefficient $h$ for air under natural convection is in range 2.5 - 25 W / (m^2 K).

> One thing that is slightly confusing to me is that heater_source = td.UniformHeatSource(structures=["heater_outer"], rate=volumetric_heat). Here the structure is set to the outer cylinder instead of the ring. Is it fine because the inner cylinder overrides the outer cylinder so it automatically knows the actual structure is a ring? 

Yes, exactly. It applies heat source only in the part of the structure not overlapped/removed by other structures

> Also one general question I have is how to pick a specific surface of a structure? For example, if I have a Cylinder inside a Box, MediumMediumInterface will be the outer surface of the cylinder. What if I want to apply certain boundary condition only on the top surface of the Cylinder?

and

> I also have another question: is it possible to define different boundary conditions to boundaries sharing the same material interfaces? For example, if the simulation did not include the bottom Si wafer, we'd have 2 SiO₂/air interfaces with different BCs. Is that possible to define using the placement argument in some other way?

Yeah, without a graphical cad interface you have to get somewhat creative and use your imagination to select particular surfaces. Typically you could create phantom structures with particular names. For example, in Lucas's example we could create structures `air_bottom` and `air_top`, then bottom and top surfaces of cladding can be specified as `StructureStructureInterface(structures=["clad", "air_bottom"]` and `StructureStructureInterface(structures=["clad", "air_top"]`, correspondingly. Or even just `StructureBoundary(structures=["air_bottom"]` and `StructureBoundary(structures=["air_top"]`. This is not ideal and I will keep looking for ways to improve that. One thing I am considering to add is to be able to point at a particular structure surface using keys `x+`, `x-`, `y+`, `side`, `base`, `top` etc. If you have any other suggestions, please let me know

> For the rings, you can use our new hole support: geometry = td.Cylinder(... outer radius ...) - td.Cylinder(... inner radius ...). This works even if there are other structures in the hole region.

Unfortunately, heat solver backend preprocessing doesn't support that yet.

> Is there any reason to create a separate heater_gap structure? Shouldn't box_clad be enough?

This is just to make the heater not a complete circle

> Is it possible to visualize the heat grid?

Not straightforwardly. Currently, you can directly download the grid "mesh.cgns" or solution on unstructured mesh "volume_proc0.vtu" and open it in, for example, paraview. Currently planning on how to make it easier to access and work with. Maybe some sort of MeshMonitor

lucas-flexcompute:
Understood. I think the BC definitions are quite flexible the way they are. Are they made like this so that you can specify non-rectangular domains? That is, if you specify an internal interface, would the solver ignore domains outside that interface?

OTOH, if the user is only specifying one of the 6 box boundaries, using (+/-) x, y, z should suffice and be easier to specify, so maybe we should support this as well?

tomflexcompute:
> * The gap between ring and waveguide is so small that it is exciting counter-propagating waves in the ring (see the standing-wave pattern in cell 97). It's also unreasonable from a fabrication point of view, so we should probably increase it to 100 nm, which is much more common. @tomflexcompute this is also a problem in the current ring resonator example, btw - I noticed that during ECOC and forgot to mention.


Thanks @lucas-flexcompute for bringing this up. Are you referring to the GUI ring resonator example?

lucas-flexcompute:
That's right. The field pattern inside the ring indicates counter
propagating waves, Wich were crated by reflections in the gap region, I
believe. I'd try increasing the gap size a little to see if the effect
disappears. (Another explanation would be reflections generated by
staircasing at the ring edges, but that would also be accompanied by
losses, which doesn't seem to be the case.)

tomflexcompute (approved):
Thanks @dbochkov-flexcompute . I think it's good to go. 

By the way, I saw the metadata info has been added to the metadata file, is the same info added to the notebook itself following this [instruction](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e)? If not, could you add it as well?

dbochkov-flexcompute:
> By the way, I saw the metadata info has been added to the metadata file, is the same info added to the notebook itself following this [instruction](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e)? If not, could you add it as well?

no, it haven't been done in other heat solver notebooks either, fixed that, thanks!

forgot to list changes after initial commit:
- got rid of `num_modes=2`
- added note on > 10 FlexCredits
- reduced simulation height in optic simulation. also removed heater and substrate from optic simulation since they might fall into pml
- switched coordinates to z=0 being at the bottom of waveguides
- fixed `Typo at "we enforcing approximately…"`
- fixed "The heat plot in cell 75 is not gray scale, as indicated in the comments and color bar. Same in cell 87."
- fixed `The creation of the perturbed optical sim with exclude={"type"} is very non-intuitive for users.`: explicitly updating structures now
- increased coupling gap between
- increased lateral size of heat simulation to make sure boundary conditions don't affect the overall results
- picked another frequency in field monitor for a nicer demonstration

lucas-flexcompute (approved):

PR #306, opened 2023-10-16
Updating estimate_cost and real_cost calls in all notebooks

momchil-flex:


tomflexcompute (approved):
Thanks. That's a great simplication!

PR #307, opened 2023-10-17
Inclusion of cavity figure-of-merit tutorial.

e-g-melo:
This is a second tutorial related to Resonance Finder where we show how to obtain the effective mode volume and Purcell factor in addition to the cavity quality factor.


caseyflex:
This will be very helpful! At some point there was discussion about including this functionality directly in the ResonanceFinder plugin. But having the code in a notebook is good -- moving it to the ResonanceFinder plugin would be a lower priority.

A few minor comments:
 
- Some of the equations are rendering incorrectly. I also think you could move all the numbered references to the top of the notebook to make them easier to find.
- Might be good to mention the Q value and V_eff reported in the paper, to give a sense of the accuracy expected
- Regarding source of error for V_eff, I wonder if the grid resolution is also an important factor here.

tomflexcompute (commented):
Thank you very much @e-g-melo for another great tutorial. Mode volume calculation has been a constant question from users and now we can finally just show them the notebook to learn from instead of explaining how to do it in text. 

Regarding the integral calculation, your calculation is totally fine. Tyler introduced me to the [integrate](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.integrate.html) method native to xarray, which we all find pretty convenient and elegant. It's also how Alec did the integral in the bistability notebook. Could you also adapt to this approach here as it looks a bit simpler? I think in general we want to educate users on xarray functionalities since they are quite powerful actually but not many users know/use them unfortunately.

e-g-melo:
Hi @tomflexcompute and @caseyflex! Thanks for all your very constructive comments. I've implemented all the changes. BTW, the xArray.Dataset.integrate() function is very convenient 👍

tomflexcompute (approved):
Thanks for the revision @e-g-melo . Looks very good. Could you just make sure all the referenced papers are purchased if not open access already?

e-g-melo:
Sure... I will merge it and purchase the papers.

PR #308, opened 2023-10-20
Adding two-photon absorption and others to api.rst

caseyflex:
This adds the following to api.rst:

NonlinearModel
TwoPhotonAbsorption
KerrNonlinearity
ContinuousWave
CustomSourceTime

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -160,8 +160,11 @@ Nonlinear
 .. autosummary::
    :toctree: _autosummary/
     
-    NonlinearSpec
-    NonlinearSusceptibility
+   NonlinearSpec
+   NonlinearModel
+   NonlinearSusceptibility
+   KerrNonlinearity
+   TwoPhotonAbsorption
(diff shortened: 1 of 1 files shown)


tylerflex:
I forget (feel free to correct me) we probably dont want to support `ContinuousWave` officially? since it doesn't ramp down I think it can have weird effects in the frequency-domain monitors? Maybe @momchil-flex remembers. If that's the case, let's leave it uncommented.

momchil-flex:
I think I'm fine with supporting it officially, just let's add to the documentation that field decay will not happen, so the user should note that the simulation will run for the full `run_time`, and the normalization of frequency-domain monitors will be arbitrary.

tylerflex:
can we merge this?

caseyflex:
After TPA is merged, frontend and backend

tylerflex:
oh right, of course 🤦 thought it was just something we forgot to merge for a second.

PR #309, opened 2023-10-20
add `signac` tutorial to ParameterScan notebook

tylerflex:
Some users were finding it challenging to run parameter scans using tidy3d, so I tested out [signac](https://signac.io) as a possible solution. This PR adds a `signac` tutorial to the `ParameterScan` notebook, which covers topics like defining parameter scans, caching, and parallel processing.  If users find difficulty in the future with managing their local storage / parameter scan results, we can direct them to this section of the notebook.


tomflexcompute (approved):
Thanks @tylerflex for the great demonstration. I found the signac workflow still a bit complex so it might take a little bit of a learning curve. Nontheless it's a very useful demo for users who would like to experiment with some advanced features.

I think "grab it's inputs" is meant to be "grab its inputs"?

tylerflex:
Thanks @tomflexcompute. yes it's a bit of a steep learning curve. Some of our users mentioned that a pain point when doing parameter scans is keeping track of all of the simulation runs and data in local storage, which signac seems to do quite well. So hopefully this at least shows them an alternative.

I fixed that typo with "it's".

PR #310, opened 2023-10-20
Apply time modulation to materials

weiliangjin2021:
- To run the notebook, please checkout `weiliang/time_modulation` in the frontend
- Above Cell 11, I'm not sure how to add link to `faq` section. Please advise.
- Also plan to add another notebook to reproduce https://www.nature.com/articles/nphoton.2008.273, but probably in another PR.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -163,6 +163,17 @@ Nonlinear
     NonlinearSpec
     NonlinearSusceptibility
 
+Time Modulation
+^^^^^^^^^^^^^^^
+
+.. autosummary::
+   :toctree: _autosummary/
+
+   ModulationSpec
+   SpaceTimeModulation   
+   ContinuousWaveTimeModulation
+   SpaceModulation
+
 Material Library
 ----------------
 
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -35,6 +35,7 @@ Mediums
     notebooks/FullyAnisotropic
     notebooks/Gyrotropic
     notebooks/CustomMediumTutorial
+    notebooks/TimeModulationTutorial.ipynb
 
 Structures
 ~~~~~~~~~~
--- a/docs/source/notebooks/TimeModulationTutorial.ipynb
+++ b/docs/source/notebooks/TimeModulationTutorial.ipynb
@@ -1,0 +1,2341 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "f7ee7d45-4de6-4492-90ce-2134a76235d5",
+   "metadata": {},
+   "source": [
+    "# Applying time modulation to materials"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a3add6f0-739b-4452-9700-eb20e31b5b2c",
+   "metadata": {},
+   "source": [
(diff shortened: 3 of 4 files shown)


tomflexcompute (commented):
Thanks @weiliangjin2021 for the implementation of the time modulation and the tutorial. Currently there are two amplitudes and the overall modulation amplitude is the product of the two, as discussed in the tutorial already. I wonder why or when it is useful to separate them? It seems to me that the time modulation part can only require frequency and phase arguments for simplicity.

e-g-melo (approved):
Hi Weiliang! This is a fascinating new feature, congrats!

I have only a few comments:
- Use capital letters in all words of h2 headers.
- Is [this](https://docs.flexcompute.com/projects/tidy3d/en/latest/faq.html#how-are-results-normalized) the link you are looking for in "??? what's the link???"?
- Add a `plt.show()` in cells [12], [15], and [24] to suppress text output.
- Before cell [18]: "complicated modulations" -> "complicated modulation"
- In notes: " at infinite frequency eps_inf, eps_inf must be positive" -> " at infinite frequency, eps_inf must be positive"
- I'd like to know if you can integrate over all the frequencies to obtain the relative photon number.
- Do you have any thoughts about why the lower sideband is suppressed in the second example?

Thanks!

tomflexcompute:
> > Thanks @weiliangjin2021 for the implementation of the time modulation and the tutorial. Currently there are two amplitudes and the overall modulation amplitude is the product of the two, as discussed in the tutorial already. I wonder why or when it is useful to separate them? It seems to me that the time modulation part can only require frequency and phase arguments for simplicity.
> 
> What you mentioned applies to phase as well: the overall modulation phase is the sum of the two?
> 
> One useful case is when the SpaceModulation is not uniform in space, so that the amplitude in time modulation can scale them together. But if it is confusing, we can consider removing it.

Thanks for the clarification. Yeah I think we just need one overall complex amplitude. I don't think the current implementation is confusing by any means. It just seems to me that it can be simplified to require fewer user inputs. Not a huge issue so we can revisit it later. Otherwise I think the notebook looks good to go. Emerson has covered pretty much all I planned to bring up.

By the way, I saw the metadata info has been added to the metadata file, is the same info added to the notebook itself following this [instruction](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e)? If not, could you add it as well?

weiliangjin2021:
> By the way, I saw the metadata info has been added to the metadata file, is the same info added to the notebook itself following this [instruction](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e)? If not, could you add it as well?

Good point, I didn't. Will follow the instruction now.

tomflexcompute (approved):
Thanks!

weiliangjin2021:
> The example where all the power gets up-converted only is great! I think there too you can add a bit more discussion though, right now it sounds a little as if it "just so happened". I think instead already when defining the simulation you could give some background on what should be expected and why? Also, do you know how the choice of the specific phase distribution affect the result? If I wanted to down-convert only, would I make the phase wind in the opposite direction, or would I make it go from 0 to pi only instead of from 0 to 2pi?

Sorry @momchil-flex somehow I edited your comment, and don't know how to restore.

I actually don't know why. I tried a few parameters, and this one looks interesting. So it's "just so happened" to me. I'll see if there is any simple explanation, if not, tunning parameters a bit in this example so that it doesn't get only up-converted.

weiliangjin2021:
> I'd like to know if you can integrate over all the frequencies to obtain the relative photon number.

A small discussion on integrated photon number has been added.

PR #311, opened 2023-10-25
Use geometry transformations

lucas-flexcompute:
Based on the changes in https://github.com/flexcompute/tidy3d/pull/1203

@tomflexcompute Do you think there are other notebooks that can use these transforms to facilitate the creation of geometries?


tomflexcompute (approved):
Thanks @lucas-flexcompute these functions are great to have! I think metalens is probably the only one that we can demonstrate rotation so far.

tylerflex:
@lucas-flexcompute do you have that original geometry transformation demo notebook? I think it could be useful to include it as a tutorial on the geometry transforms eventually if it's not too much effort to clean it up.

lucas-flexcompute:
It's not really a demo, more of a test notebook that I used for debug and to create the actual tests: [here it is](https://gist.github.com/lucas-flexcompute/da6c723e265078602ef7b6e51792932b)

PR #312, opened 2023-10-26
Added radiative cooling example notebook and associated files

tomflexcompute:
This example is contributed by Zongfu's postdoc and is associated with a paper that will be published soon. The script was written by the contributor and largely unchanged. I mainly organized it a bit and added comments and markdown explanations. 

We need to publish this example by tomorrow so your timely review is much appreciated. Thanks @e-g-melo @caseyflex !


e-g-melo (approved):
Hi Tom! This example is very nice, it shows an FDTD application pretty different from the ones we are used to seeing. My comments follows next:

1 - What is the purpose of `lda_start = wl_start` in cell [2]. It seems that only one of these variables is enough.
2 - It would be interesting to include another cell to fit the SiO2 and Al2O3 materials so the readers can easily follow the code and results.
3 - Messages like "print('Geometry Finish')" would be helpful in a Python script, but it seems to include not useful information in a notebook.
4 - The grid size (`td.GridSpec.uniform(dl =lda_start/10)`) of 400 nm seems too big compared to the Al2O3 size. If this value is enough in this application, maybe it would be nice to include a note about this.
5 - I wonder about the purpose of adding `time.sleep(4)` in the cell [9].

caseyflex (approved):
Very cool example and nice notebook!

I'd echo Emerson's question about the grid resolution. Also, I wasn't sure if we could provide some intuition for the final results plots? Why is there a dip in absorption in a certain range, and why is the field localized in a certain region? Or is this analysis something saved for the paper.

Otherwise, looks good to me. Thanks!

tomflexcompute:
Thanks both for the prompt review on short notice and the great comments. To answer Emerson's questions:

1. Yeah it seems like there are a few variables that are defined but not really used. Maybe the author was experimenting with something but didn't end up using it and forgot to delete it either. I just cleaned them up.
2. Done.
3. Indeed. These printings have been removed. Right now only the number of particles is printed, which I think might be useful information to see.
4. I personally also think it's a bit large. I reduced it by half now. 
5. The sleep time is artificially added here to give the server some buffer time. Otherwise the real cost will be returned as 0 since the cost calculation takes a few seconds. Turns out that 4 seconds might not be enough in some cases. To ensure we get the real cost correctly, I increased the sleep time to 20 seconds just to be safe.

And to answer Casey's question: I'm not particularly familiar with the detailed mechanism behind it either. Will add the reference and more information about it once their paper is published online.

PR #313, opened 2023-10-26
adjoint circuit demo with sax

tylerflex:
combines two tidy3d adjoint simulations with a circuit simulator (sax) into a demo optimizing an MZI at the circuit parameter level while co-optimizing the individual simulations automatically.

Note: am re-running right now for better results, but the main notebook will be like this.


tylerflex:
just re-ran notebook and added a figure
image

tomflexcompute (approved):
Thanks Tom

> 1. The klujax warning in the beginning seems a bit annoying. Do you want to get rid of it? Would it go away after installing klujax?

I tried installing this but it didn't seem to work. So I'm thinking I might just leave it.

> 2. From the mode solving we see that the waveguide at this particular width only supports two modes. The last two modes with effective indices around 1 are spurious. Maybe we can only solve 2 modes in this case?

I reduce the `num_modes` in the `ModeSpec` and added an explanation to the user that we only want to solve up to the modes of interest to save computation. But for the initial mode solve (for setting the mode index and mode spec) I left all of the spurious modes just as a demonstration. I think this should be ok and maybe instructive? 

> 3. It seems like the notebook is not ran from the beginning to the end in one run so the cell number is discontinuous. Could you rerun it before merging?

I would but it takes several hours to run and I don't think it would be good to burn credits. Perhaps we can leave it until the next time we need to run notebooks unless you think it's too distracting for the user.

> 4. Thanks for adding the metadata in the metadata.txt file. Could you also copy/paste it into the notebook? I can do so and push the commit if you prefer. 

Hm, are there some notes on where this should go? I'm not sure where to put it, maybe you can make a commit and I'll look at the diff for next time?

> Some typos and potential wording issues:

guess my jupyter spell checker plugin stopped working, thanks for catching these. All fixed.

> 1. "indivual" -> "individual"
> 2. "The scattering matrix of the two components" -> "The scattering matrices of the two components"? Or "are" -> "is"
> 3. "funciton" -> "function"
> 4. "intput" -> "input"
> 5. "inect" -> "enter"?
> 6. "make sure it worked properly" -> "make sure it works properly"?
> 7. "as in in the other examples" -> "as in the other examples"

I pushed a new version, let me know if you think it's necessary to re-run the whole notebook or ok to leave it, also could you add the metadata or direct me to the documentation explaining where that should go? (if it's not written down somewhere, even in the wiki of this repo then maybe we should add it).  thanks!

e-g-melo (approved):
Hi Tyler! This notebook is fantastic, congrats! I think you should plan a publication if not already planned :)

I have only some minor comments on it.
1) Why do we need `pip install "tidy3d[jax]"` in this case? What is the difference to installing `jax` and `tidy3d` individually?
2) Why do we have this `-1e-5` in some coordinate definitions to `JaxDataArray`?
```
    coords_x = [x0_min + index_x * size_box_x - 1e-5 for index_x in range(nx)]
    coords_y = [y0_min + index_y * size_box_y - 1e-5 for index_y in range(ny)]

```
2) `k0` and `eps_max` are defined but not used, and `BinaryProjector` is imported but not used.
4) In addition to Tom's comment on mode solver, I suggest showing only the y and z field components to reduce the number of plots.
5) Before cell [27], why is the maximum possible objective value `0.0`? It that a typo? The image shows ~0.5. 
6) It would be nice to delete the commented line in cel [38].

Thanks!

tylerflex:
Thanks Emerson

> 1. Why do we need `pip install "tidy3d[jax]"` in this case? What is the difference to installing `jax` and `tidy3d` individually?

One doesn't need to `pip install "tidy3d[jax]"` but it will install the proper supported versions for windows vs. unix and also install `jaxlib` which one needs to install first. So it just makes the process simpler for users who might not be comfortable doing it themselves.

> 2. Why do we have this `-1e-5` in some coordinate definitions to `JaxDataArray`?
> 
> ```
>     coords_x = [x0_min + index_x * size_box_x - 1e-5 for index_x in range(nx)]
>     coords_y = [y0_min + index_y * size_box_y - 1e-5 for index_y in range(ny)]
> ```

I think this was for historical reasons and probably isn't important anymore. My thought was that this was to avoid an issue if the coordinate was right on a grid cell boundary. In that case, the numerical precision could affect whether one grid cell or another was assigned the permittivity values. I think this 1e-5 makes it explicit that the grid cell to the - side of the boundary is assigned. Perhaps we can remove this though.

>  2. `k0` and `eps_max` are defined but not used, and `BinaryProjector` is imported but not used.

Thanks, I'll remove them.

> 3. In addition to Tom's comment on mode solver, I suggest showing only the y and z field components to reduce the number of plots.

Thanks, removed Ex.

> 4. Before cell [27], why is the maximum possible objective value `0.0`? It that a typo? The image shows ~0.5.

That was a typo from before, removed it.

> 5. It would be nice to delete the commented line in cel [38].

Removed. 

Thanks!

tylerflex:
@tomflexcompute I added the metadata to the notebook, for some reason I thought it was more complicated than that :). I'll try to remember to add it both times in the future, or we could try inserting it from metadata.json through a script

tomflexcompute:
> @tomflexcompute I added the metadata to the notebook, for some reason I thought it was more complicated than that :). I'll try to remember to add it both times in the future, or we could try inserting it from metadata.json through a script

Thanks! Actually there is a script update_notebook_metadata.py that inserts/updates all notebooks with the metadata in the metadata.txt file. However running this script will change some metadata like version number etc even though the metadata contents are unchanged. It might be fine but you can also modify the script to only insert/update metadata to certain notebooks.

image

PR #319, opened 2023-11-01
Inclusion of adjoint light extractor.

e-g-melo:
The notebook describes the optimization of a quantum emitter light extractor. I've used `FieldMonitors` to calculate the dipole power and get the coupling efficiency from the dipole into the output waveguide. The result seems fine, but I will run it one more time trying to improve the final CE.


tylerflex:
Thanks @e-g-melo this is a really cool demo and the final structure is very nice. Just a few comments, mainly wondering about the final objective function plot.

> The QE is placed within the design region so that we will surround it with a constant refractive index region protected from etching.

"The QE is placed within the design region so we surround it with a constant refractive index region to protect it from etching." (Double check whether that makes sense here).

>     eps **=** jnp**.**where(eps **<** eps_min, eps_min, eps)
>     eps **=** jnp**.**where(eps **>** eps_max, eps_max, eps)

If it’s useful, a differentiable version of this is
```py
eps = jnp.maximum(eps, eps_min)
eps = jnp.minimum(eps, eps_max)
```

Let’s add a note explaining why we see the warning in cell [9] about structure overlap and why it’s ok to ignore.

In `def penalty`, maybe add a docstring or note in the markdown above saying that this penalizes if the structure isn’t invariant under erosion and dilation, which signifies small feature sizes. 

Can you explain why the objective function seems to decrease for most of the optimization steps? Is the learning rate too large? Maybe you are running this again.

momchil-flex:
I got curious about this so I had a look. 

- I have a similar question to Tyler's

> Can you explain why the objective function seems to decrease for most of the optimization steps? Is the learning rate too large? Maybe you are running this again.

image

It does seem to me that the learning rate may be too large? It seems like the device jumps to the highest coupling efficiency right away, and the rest of the optimization is just the fabrication penalty going down?

- In the overlay in the last image, the symmetry is not properly applied for some reason:
image

- I think this code

```
qe_field_plan = [
    td.Box(center=(cr_center_x + cr_l / 2, 0, 0), size=(0, cr_w, 2 * wg_thick))
]
qe_field_plan += [
    td.Box(center=(cr_center_x - cr_l / 2, 0, 0), size=(0, cr_w, 2 * wg_thick))
]
qe_field_plan += [
    td.Box(center=(cr_center_x, cr_w / 2, 0), size=(cr_l, 0, 2 * wg_thick))
]
qe_field_plan += [
    td.Box(center=(cr_center_x, -cr_w / 2, 0), size=(cr_l, 0, 2 * wg_thick))
]
qe_field_plan += [td.Box(center=(cr_center_x, 0, wg_thick), size=(cr_l, cr_w, 0))]
qe_field_plan += [td.Box(center=(cr_center_x, 0, -wg_thick), size=(cr_l, cr_w, 0))]
```

can be replaced with something like `qe_field_plan = td.Box.surfaces(center=(cr_center_x, 0, 0), size=(cr_l, cr_w, 2 * wg_thick))`.

e-g-melo:
Hi @tylerflex and @momchil-flex! Thanks for reviewing the notebook and sending your comments. I will replace the codes you have pointed out and fix the texts. Regarding the decreasing coupling efficiency, I guess the coupling to radiation modes is sensitive to fabrication penalty constraints. I did some optimization where I applied the fabrication penalty right at the beginning of the optimization and the convergence was very bad, resulting in lower coupling efficiencies. So, including the penalty function after some iterations I've obtained a better result. I am trying to change the point in which to include the penalty. Hopefully, we will have a better result.

tomflexcompute (approved):
Another great example thanks to @e-g-melo ! Looks very good already. Some comments below:

The initial jump from <0.2 coupling efficiency to ~0.7 in one iteration is indeed quite interesting. I wonder how the device looks like at that stage. 

Does this optimization cost over 10 credits? Could you add a note at the beginning to warn users about the cost roughly speaking so they can develop a better sense on how expensive a typical optimization run will cost?

The final design as you stated looks like a nanobeam cavity. Maybe add a sentence or two on the physics behind it? Such as the holes on the right function like a bragg mirror to reflect the radiation to the -x direction.

Currently the design region has the same width as the waveguide. I wonder if making it a bit wider, say 1 um, would make the optimization easier and performance better without much increase in computational cost. Nonetheless, it's just a thought. As a demo example, we don't need to nitpick these details.

A typo "funciton" in the comment of cell[17]. Probably copied from Tyler's notebook :wink: ?

tylerflex:
Actually, it kind of makes sense to me now why the objective function is jumping around like this. I didn't realize that the penalty was beta-dependent. as the optimization goes on, the penalty seems to get stronger and the extraction FOM is less prioritized. I've noticed if you adjust the objective function in this way, the final plots can look a bit strange like this, perhaps it's ok if we mention something in a note right after this figure.  Or try a scheme where the penalty is constant w.r.t. iteration number

tomflexcompute:
Very nice results! Thanks Emerson. 

Could you add the imported file name to the `import_file_mapping.json` file in the misc folder?

e-g-melo:
Done! @tomflexcompute, what is the purpose of this file?

tomflexcompute:
> Done! @tomflexcompute, what is the purpose of this file?

Thanks Emerson. This file essentially tells MC which files in the misc folder need to be zipped with the corresponding notebook file. When a user wants to run this notebook in our hosted notebook environment, the imported files will be added so the notebook can be run correctly.

e-g-melo:
@tylerflex and @momchil-flex, can I merge this PR?

tylerflex:
Fine by me, thanks Emerson

momchil-flex (approved):
Yep looks great!

PR #320, opened 2023-11-01
Added xarray tutorial

tomflexcompute:
We have a [tutorial](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/VizData.html) on basic postprocessing and visualization already but from the customer support requests, a fair amount of users still have questions about xarray manipulations. Hopefully this more in-depth tutorial can help them (mainly targeting users who have used tidy3d for a while but are not very comfortable with xarray manipulations). Feel free to suggest anything else that you think is worth including.


tylerflex (changes requested):
Thanks @tomflexcompute I think this will be extremely useful!

>  As discussed, in Tidy3D, monitor data are [[xarray.DataArray](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html#xarray.DataArray)](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html#xarray.DataArray) objects

I think this is a little misleading because `td.MonitorData` are not `xarray.DataArray` but contain them. Probably best to re-phrase as “As discussed, in Tidy3D, we represent multi-dimensional data (such as arrays) using `xarray.DataArray` objects.”

>  such as performing interpolation and integral.

“Such as performing interpolation and integration.”

> We can access each monitor data by specifying the key. 

I’d probably be more specific and say “We can access the data for each monitor by indexing into the `SimulationData` with the monitor `.name`.”

Sometime soon after,  it might be useful to talk about why we do `.flux` for example.

“The monitor dataset contains one or more `xarray.DataArray` objects storing the actual raw data. For the flux monitor data, we can access the raw flux data as a function of frequency with `.flux`.”

I think it’s also useful to point out here that the data has `.values` and `.coords`. The `.values` are just the raw, unlabelled data. The `coords` are a dictionary mapping each dimension of the data to a set of data. For example, mapping the flux values to a specific frequency.

Just FYI, the `.abs` is something we added to our custom `DataArray` subclasses, but doesn’t exist in `xarray`. I’d recommend switching to use `abs()` instead, which operates the same way and is supported by `xarray`.

It might be also worth noting that standard mathematical operations (`+`, `-`, `*`, `**`) are supported as expected.

More of an advanced use, but note that one can supply `slice()` to the coordinates in eg. `sel` so I think `obj.sel(x=slice(1,5,2)` will select every `x` from 1 to 5 in increments of 2. But might want to double check.

> The field data are natively located at the Yee grid

I think this changed as of 2.4.0 and they are all located, so maybe just leave it out.

Consider copying the links to the xarray documentation and video to the introduction paragraph as well, as users might want to just jump into those directly. “If you want to check out the xarray documentation directly, we recommend you check out …”

momchil-flex:
> I’d probably be more specific and say “We can access the data for each monitor by indexing into the SimulationData with the monitor .name.”

Maybe even print `sim_data.simulation.monitors` so that it's visible that there's a key for every monitor?

momchil-flex (approved):

tomflexcompute:
Thanks both @momchil-flex @tylerflex for the comments. Revisions have been implemented. 

Does supplying slice() to the coordinates in sel effectively achieve the same result as interpolate? In a sense interpolation is even better as it can be linear, cubic, etc but sel can only be picking data in the actual dataset. 

Printing `sim_data.simulation.monitors` is actually quite messy so I avoided it. The readability is not particularly good, see the screenshot below. 

image

tylerflex:
> Does supplying slice() to the coordinates in sel effectively achieve the same result as interpolate? In a sense interpolation is even better as it can be linear, cubic, etc but sel can only be picking data in the actual dataset.

interpolation is fine, slice is mainly used to apply min and max bounds and also steps. So `slice(x, y)` will select the data between x and y. `slice(x, y, 2)` will select every 2nd data point. So I think the use case is more restricted than interp but maybe it's useful at times if you want to restrict to selecting the actual data. Feel free to leave this out though.

tomflexcompute:
> > Does supplying slice() to the coordinates in sel effectively achieve the same result as interpolate? In a sense interpolation is even better as it can be linear, cubic, etc but sel can only be picking data in the actual dataset.
> 
> interpolation is fine, slice is mainly used to apply min and max bounds and also steps. So `slice(x, y)` will select the data between x and y. `slice(x, y, 2)` will select every 2nd data point. So I think the use case is more restricted than interp but maybe it's useful at times if you want to restrict to selecting the actual data. Feel free to leave this out though.

I see. Indeed I think it's a less commonly used function. We can leave this out for now but if we see users showing interest in this we can add it later on.

PR #322, opened 2023-11-07
Inclusion of CMOS RGB image sensor.

e-g-melo:
Hi! I've included the fitting of RGB materials using the FastDispersionFitter. The optical efficiency is well-defined now and better than the ones reported by the Lumerical example.


tomflexcompute (changes requested):
A great example that has been long awaited by many industrial users! Thanks @e-g-melo ! The performance of the sensor also looks very nice.

Some minor comments for your consideration:

1. In the note, FlexUnit should be FlexCredit. I suppose this was written a while back when we were still using FlexUnit.

2. "The height of len is 0.339 micron" probably meant to be "The height of the lens is 0.339 microns".

3. "let's look the obtained refractive index and extinction spectrum" -> "let's look at the obtained refractive indices and extinction spectra".

4. "plane wave source at normal direction" might be better phrased as "plane wave source at normal incidence".

5. In the introduction, "Tidy3d" -> "Tidy3D".

6. The fit of the real part seems quite off, which is likely due to the small weight. Maybe you have tried this already but is it possible to have some `weights` value such that the real part also fits a bit better?

7. The `run_time` seems a bit long since the field decay reached the threshold at 28%. Since we charge at least 1/3 of the credits, we generally want to keep the shutoff above 33%. In this case, maybe we can reduce `run_time` by a factor of 2?

8. I noticed that you gave all mediums and structures a name, which is fine since we do have this argument. However it's not necessary in general. When you define a structure, it's ok to just give it a name since it's just an extra argument. However, naming the medium with the extra copy commend complicates things unnecessarily in my opinion. Do you have other reasons why it might be beneficial to do so?
image

e-g-melo:
Thanks for your comments @tomflexcompute!

Regarding 2: I tried other values of weights, but they didn't help to obtain a good compromise between the real and imaginary part. See below:

**weights=(0.2, 1.8)**
image

**weights=(0.3, 1.7)**
image

**weights=(1.5, 0.5)**
image

The material rename doesn't seem necessary. I've excluded them. @xin-flex, as you have coded the notebook first, let me know if you have any particular idea in mind about that. 

I've corrected the other items.

tomflexcompute:
Thanks @e-g-melo for the clarification. I suppose the artificial refractive index is too far from Kramers–Kronig relation compliance so it's not possible to fit both accurately at the same time. It's fine. Maybe just add a comment explaining why the fit is not good but it's ok for our simulation?

e-g-melo:
Nice, I will include a comment there.

weiliangjin2021:
The color filter material now looks so nice, so much better than the time I tried to fit them! Regarding the lens, do you think it's worth mentioning that we can have ellipsoidal shape with the transformation method?

e-g-melo:
Hi @weiliangjin2021! Thanks for this comment. Actually, it would be a perfect opportunity to try the transformation. I will see try to include it in the notebook.

e-g-melo:
@weiliangjin2021, I've used the `scaled` method to transform the sphere in an ellipsoid, and the optical efficiency improved slightly. However, the transformed structure didn't render when using `plot_3d()` (which I included to show the 3D sensor image). I think we can make the notebook available in its current state and include some transformations afterward publishing a tutorial on these new transform features included in 2.5.0rc2.

xin-flex (approved):

tomflexcompute:
Looks very good Emerson! Feel free to squash and rebase/merge.

PR #321, opened 2023-11-06
Merge `tidy3d` and `docs` 

daquinteroflex:
See more information in https://github.com/flexcompute/tidy3d/issues/1231

PR #323, opened 2023-11-08
Convert to notebooks only submodule repo

daquinteroflex:
See more in https://github.com/flexcompute/tidy3d/issues/1231

We've decided we want to convert this repository into a notebooks only one. So we're closing trying to merge the repositories based on history https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/321, and just converting the structure of the repositories to just contain a notebooks submodule. This should streamline the documentation management and usage in a way that aligns currently with what we're trying to do.


daquinteroflex:
The way I believe this repository should work is completely standalone. It contains only application examples, and should just need to be configured from a wiki. The only requirement is the current `tidy3d` project module.

docs/requirements.txt
tylerflex, line 1: Just a note, many of the example notebooks introduce additional dependencies (such as `tmm`, `sax`, etc) that aren't a requirement for the frontend itself or for building the docs. So I think we could consider storing these somewhere.

tylerflex: to clarify, these requirements are because they are used in the notebook examples themselves. to support the demo.

tylerflex:
> The way I believe this repository should work is completely standalone. It contains only application examples, and should just need to be configured similar as a wiki. The only requirement is the current tidy3d project module.

So this repository will be a submodule in `tidy3d`? if so, how do we also add `tidy3d` as a requirement? isn't it recursive? or am I not understanding something. If completely standalone how will the `tidy3d/docs` reference it, via URL? Maybe you can explain briefly so I understand better.

> Someone should just want to say, make an example: they create a notebook and add it into some relevant directory. Based on the name of the example, this gets compiled into an index.rst file that gets built both on the web and on the documentation. You want to have all the tools installed and set up here so you can easily merge and integrate changes. In my opinion though, those tools should be installed from the pyproject.toml file as a way to configure the full development environment for the user.

I think the main use case is that someone is interested in reading or trying out some examples of tidy3d, so priority 1 should be making these examples very easy to find, display, and run either locally or in the cloud. So maybe binder or some kind of other tools for displaying / running notebook only repositories could be useful here.

If a user wants to contribute an example, I think we can review it here and then configure the `tidy3d/docs/examples.rst` to add it if we want. I don't think we want to automatically add any user-generated examples to our docs. but maybe if we create a separate section in the docs for these, we can think about that. It might be unlikely that we get such contributions, but who knows.

On that note, if you want to contribute the tidy3d demo notebook you made to learn more about tidy3d, feel free to make a PR for that!

I think separating dependencies makes sense. There are really 3 general sets of dependencies relevant to this I think:
1. `tidy3d` dependencies (everything needed to run `tidy3d`) (pydantic, numpy etc)
2. `docs` dependencies (everything needed to build the docs) (sphinx, etc.)
3. `notebook` dependencies (extras that are used in these notebooks) (`tmm` etc)

After merging we will store configuration for 1,2 in the front end `tidy3d` and 3 in this repo? What I had in mind was that this repo would basically just be a directory of `ipynb` files and some `requirements.txt` (or equivalent) with the extra requirements.

docs/requirements.txt
daquinteroflex: Just saw, so this is structuring my thoughts of how this could work, again we don't have to follow it, it's mainly to get an idea of how/why something gets used and how to improve the workflow for that. I guess I began thinking about this because currently we're still thinking of submodules, but I was trying to understand why I felt this made more sense. 

In summary, all dependencies for docs/notebooks/source code get currently managed from `pyproject.toml`. This means that they only have to be able to install the package in a given option, say `tidy3d[dev]` to manage all of this, and that way it does not get managed separately in this repo.

daquinteroflex:
Will just summarize what Tyler and I were chatting in the slack for posterity:

- `pip install tidy3d[notebooks]` is everything people have to run to use all these notebooks. Managed from `pyproject.toml`
- This repo becomes a directory of examples that can be cloned and easily run. 
- Think about binder integration, where does it get configured

PR #325, opened 2023-11-18
Inclusion of fabrication constraint in level set notebook.

e-g-melo:
Inclusion of the minimum gap and radius of curvature constraints (Scientific Reports 9, 8999 2019 DOI: https://doi.org/10.1038/s41598-019-45026-0) in the level set notebook.


tylerflex (changes requested):
This is fantastic to include, thanks @e-g-melo. I think this really strengthens our docs and also makes this method much more accessible (and more likely to be used generally).

A couple typos / rewordings.

[10] control knots -> control knobs.
[24] The coupling into the output waveguide _improves fast_ in the beginning at the expense of higher penalty values. -> improves quickly.

Another comment, maybe consider plotting the objective function over iteration in log scale in y since it seems to decrease a lot at the very beginning

Otherwise, looks great.

momchil-flex:
Wow, I agree with Tyler, this is pretty awesome to have!

When you talk about coupling efficiency, "The resulting structure shows good performance, presenting coupling efficiencies close to -3 dB at the central wavelength.", it makes it sound like the device still has a 3dB insertion loss. In fact, it has really low insertion loss, and 3dB is just because you're looking at the power in one of the arms only. I would suggest rephrasing that. For example, you could plot the total power coming out in the two arms (we know that the device is symmetric), or normalize the power in one arm by the target value of 1/2.

I know that you guys have some thoughts about this already, but making the transition from the design region to the waveguides smooth would definitely be one last nice improvement to this notebook. The optimized design is already super nice and smooth, but there are the connection kinks, which even seem to localize field a little
image

tylerflex:
Yea we should test out some solutions to this in our existing notebooks. Related: it is pretty cool that we now have a testing bed of like 10 different inverse design problems to try out random ideas on.

tomflexcompute (approved):
Very nice results! Thanks @e-g-melo for the continuous efforts. 

A few things I caught while reading it:

1. "keep the output waveguides apart each other" -> "apart from each other"

2. "objective funciton value" -> "function"

PR #326, opened 2023-11-21
design space exploration plugin

tylerflex:


lucas-flexcompute (approved):
Just minor comments: 

In the ParameterScan notebook:
- `npts_bend` is not used in `make coupler`.
- (intro): link the Sweep notebook.
- I think it's better to explain that the Variable names must match the function arguments, in case the user does not go through the Sweep notebook.
- The coupler gap range in the sweep should probably be from 0.1 to 0.15.
- `project = signac.init_project(path="data/coupler")` to be consistent (not "splitter").

In the Sweep notebook:
- typo: "psot processing"
- I like the idea of naming outputs. Maybe use that in both notebooks to indicate to the user that it is the preferable method to facilitate operating on the data frame later?
- Just an idea: instead of the hex binning plot sample, how about showing that outputs can also be used as axis: `results.scatter(x="t", y="output", s=100, c="n", cmap="magma", ax=ax2)`

tylerflex:
Thanks @lucas-flexcompute , I implemented all of these suggestions and am re-running now. I was unclear about this one. Could you explain?

> Just an idea: instead of the hex binning plot sample, how about showing that outputs can also be used as axis:

lucas-flexcompute:
> Thanks @lucas-flexcompute , I implemented all of these suggestions and am re-running now. I was unclear about this one. Could you explain?
> 
> > Just an idea: instead of the hex binning plot sample, how about showing that outputs can also be used as axis:

In the example (and the illustration figure) you use the design variables to give values to the x and y coordinates of the plot. That's probably the most usual type of visualization, but you can also use "output" as the y value, for example, and some variable as color or size in the scatter plot. Not all users will be aware of that, and it can be an insightful look into the data.

tylerflex:
ok I understand, thanks. That's a good idea. I implemented that in the hexbin plot with a comment about it.

tylerflex:
closing as incorporated in https://github.com/flexcompute/tidy3d/pull/1199 and https://github.com/flexcompute/tidy3d-notebooks/pull/new/tyler/design/init

PR #327, opened 2023-11-28
Tom/visible edge coupler

tomflexcompute:
Our current examples are heavily biased towards silicon-based devices at telecom. A few users have expressed strong interest in nitride devices at visible, hence this example. I think at some point you shared that someone asked about examples similar to this before @momchil-flex . The size of the device is pretty large so the cost is inevitably a bit high, but it's a good highlight of our capabilities.


e-g-melo (approved):
Thanks for this great notebook, Tom! Indeed, we need more examples on shorter wavelengths.

I have only a few minor comments on it:
1) The is adapted from - > The example is adapted from
2) In [4], you wrote: "The variables are schematically shown below." Were you planning to include a schematic image like the one in Fig. 1(c) of the paper?
3) It would be an excellent opportunity to include a reference to the symmetry tutorial (https://www.flexcompute.com/tidy3d/examples/notebooks/Symmetry/) in [7]: " This polarization corresponds to symmetry=(0,-1,0)".
4) In the field plot, it would be nice to include "TE" and "TM" in the figure titles to identify them quickly.
5) The paper shows the coupling efficiency as a percentage and the coupling loss as positive values (-10Log10(T)). Should we use positive values in the insertion loss graph as well?

tomflexcompute:
Thanks @e-g-melo .These are all great suggestions! Modifications have been implemented accordingly.

Originally I planed to draw another schematic to indicate the parameters but after a second thought, it might not be necessary since the paper contains it already and readers can easily access this open-access paper. In the notebook I just reference Figure 1 in the paper.

momchil-flex:
Looks great! I think it's probably fine as is; as a note, though, when I was discussing such a setup with a user, they wanted to come in the opposite way, with a mode source in the bottom waveguide, and characterize the near and far field of the out-coming beam at the facet. I think doing this could also be a way to more precisely compute coupling to your fiber, if you know exactly what the mode looks like. In principle, if we come in this way (from the right in your setup) we could just assume a gaussian fiber mode again, and put a near field monitor at the output and compute the mode overlap to get the coupling efficiency, which should be very similar to what you get here. And we could show that the far field can be obtained too. But I still think it's fine as it is too.

tomflexcompute:
> Looks great! I think it's probably fine as is; as a note, though, when I was discussing such a setup with a user, they wanted to come in the opposite way, with a mode source in the bottom waveguide, and characterize the near and far field of the out-coming beam at the facet. I think doing this could also be a way to more precisely compute coupling to your fiber, if you know exactly what the mode looks like. In principle, if we come in this way (from the right in your setup) we could just assume a gaussian fiber mode again, and put a near field monitor at the output and compute the mode overlap to get the coupling efficiency, which should be very similar to what you get here. And we could show that the far field can be obtained too. But I still think it's fine as it is too.

That's a great idea too. I can try it a bit to see how well it works. In principle one can also just include the real fiber core in the simulation, which I kind of wanted to do but I just don't have a good set of parameters for it. @e-g-melo do you know what would be the good core radius and n_core and n_clad for the fiber?

e-g-melo:
The fiber datasheet specifies a core diameter of 3 micros and NA of 0.12. So, if you consider an oxide core at visible wavelength, I think you can obtain an estimation of the cladding refractive index.

image

tomflexcompute:
> The fiber datasheet specifies a core diameter of 3 micros and NA of 0.12. So, if you consider an oxide core at visible wavelength, I think you can obtain an estimation of the cladding refractive index.
> 
> image

Great info! Will give it a try and see how it works.

e-g-melo:
Pretty good!

momchil-flex (approved):

momchil-flex:
Nice - just squash before merging?

tomflexcompute:
Thanks guys for reviewing!

PR #331, opened 2023-12-06
Tom/tkinter gui

tomflexcompute:
So at IPC, a user brought up that the simulation software he is using allows him to build simple GUIs that he can pass on to colleagues with little simulation experience. We argued that you can do the same with Tidy3D since it's fully Python based. To make the case I did a simple demo and this is the slightly cleaned up version. Maybe there are other users curious about this possibility. I put it in a notebook as all our examples are in notebooks but not sure if that's the best way to present it.


tylerflex:
Hey @tomflexcompute , this is super cool! I think a notebook is probably the best place for it for now, I don't see a straightforward way to include this in our source code (maybe a GUI-builder plugin? But that would require a substantial effort I think).

I had a few comments, the writing is great but wanted to point out:
* I feel like a title to capture the subject might be better. For example: "Building a Simple, Python-Based GUI for Tidy3D"? I feel the application is less relevant but maybe it can be included? Just think it should emphasize the GUI building.
* Not sure about the use of `global` in the code. Is this how it's recommended to be done in tkinter? Generally this is not advised in python if possible. Any simple workarounds? if it's standard, then ignore this.
* Please add `tkinter` to the `requirements.txt` in `tidy3d-docs`.

tylerflex (approved):

tomflexcompute:
> Hey @tomflexcompute , this is super cool! I think a notebook is probably the best place for it for now, I don't see a straightforward way to include this in our source code (maybe a GUI-builder plugin? But that would require a substantial effort I think).
> 
> I had a few comments, the writing is great but wanted to point out:
> 
> * I feel like a title to capture the subject might be better. For example: "Building a Simple, Python-Based GUI for Tidy3D"? I feel the application is less relevant but maybe it can be included? Just think it should emphasize the GUI building.
> * Not sure about the use of `global` in the code. Is this how it's recommended to be done in tkinter? Generally this is not advised in python if possible. Any simple workarounds? if it's standard, then ignore this.
> * Please add `tkinter` to the `requirements.txt` in `tidy3d-docs`.

Thanks @tylerflex for the great suggestions. All have been implemented. Originally I just used global variables to save time and avoid passing variables between functions, but you are right that it's generally not advisable to use it. Now I've avoided it.

tomflexcompute:
@momchil-flex do you have any last minute comments before it's merged?

momchil-flex:
Looks good!

momchil-flex (approved):

PR #332, opened 2023-12-08
Added chiral metasurface notebook and associated files

tomflexcompute:
By popular request, we include this chiral metasurface notebook to demonstrate how to simulate circular dichroism and circularly polarized plane waves. 

In addition to what we demonstrate in this notebook, the original paper also demonstrates a coupled thermal-optics simulation where the incident light induces local heating, which leads to the phase change of GST. In our current heat solver, is it possible to simulate this @dbochkov-flexcompute , where the heat source is the absorption of electromagnetic waves?


dbochkov-flexcompute:
> In addition to what we demonstrate in this notebook, the original paper also demonstrates a coupled thermal-optics simulation where the incident light induces local heating, which leads to the phase change of GST. In our current heat solver, is it possible to simulate this @dbochkov-flexcompute , where the heat source is the absorption of electromagnetic waves?

Currently, not, but it is on our roadmap

dbochkov-flexcompute:
The notebook looks great to me, don't have anything to add. Besides maybe adding a figure similar to Fig. 1 from the paper, but leave it up to you.

And, yeah, definitely it would be nice to demonstrate different temperature responses once we add EM heating to our solver

e-g-melo (approved):
Hi Tom, thanks for this notebook! It is pretty interesting and different from the ones we have so far.

I have no other comments to add here as well. It is good to go in my opinion.

PR #337, opened 2023-12-19
Added scale-invariant waveguide notebook and associated files

tomflexcompute:
This notebook reproduces a recent [paper](https://www.nature.com/articles/s41467-023-42234-1#:~:text=The%20scale%2Dinvariant%20waveguide%20shows,distribution%20in%20the%20strip%20waveguide.). The authors have agreed to collaborate with us to do a co-marketing activity so will publish the notebook first. 

The notebook only demonstrates mode solver and waveguide plugin. The result is very sensitive to the setup so it can serve as a good test for our mode solver in the future.


e-g-melo (approved):
Hi Tom, it is an exciting work! Both paper and notebook are very nice. They have potential to create high impact.

I have no comments, only some typos:

1) After [4]: "refines a the Structures" -> defines the Structure?
2) After [4]: "The run_time is irrelevant so can be set to an arbitary value" -> The run_time is irrelevant so it can be set to an arbitrary value
3) After [4]: It seems something is missing in these sentences: "Since we are solving for the fundamental quasi-TE mode and the waveguide geometry is symmetric, we can set symmetry=(-1,1,0). This is not necessary since mode solving is usually a light computation." Maybe a "however"?
4) Before [11]: "which is a conveneient tool" -> "which is a convenient tool"

tomflexcompute:
Thanks @e-g-melo ! Somehow the typo checker stopped working today. I ran the checker again and fixed all the typos.

PR #340, opened 2023-12-22
Fixed typo in BistablePCCavity.ipynb

alec-flexcompute:


tylerflex:
@alec-flexcompute in the future it's best to get a review before merging a PR, even if it's just a minor one like this. The reason is we have a 2nd pair of eyes in case something sneaks into the PR that can become an issue down the line. (It happens more often than you'd think).

It also might be good to fix this in the upcoming pre/2.6. maybe @tomflexcompute  can help there.

alec-flexcompute:
Gotcha, sorry about that. And yes I'll change it in pre/2.6 with Tom as
reviewer.

PR #346, opened 2024-01-12
A few corrections in heat solver notebooks

dbochkov-flexcompute:
- mentioned `pip install tidy3d[vtk]`
- fixed typo "50 mA" -> "30 mA"
- also, noticed that some thermal quantities were not given in the right units (results were right though)

ignore sudden change in mode solver results in thermal ring resonator notebook, this was happening before as well and it seems to be due to using local mode solver. Starting 2.6, it is switched to remote mode solver (it required new `Simulation.subsection functionality`)


dbochkov-flexcompute:
once this is merged, I'll do the same updates for new 2.6 docs

tomflexcompute (approved):
Thanks @dbochkov-flexcompute !

PR #347, opened 2024-01-24
Update import_file_mapping.json

tomflexcompute:
MC caught this missing file in the import map. Will make another PR for the same change in the tidy3d-notebooks repo after this merge.


e-g-melo (approved):
Good catch!

PR #348, opened 2024-02-05
changed nonlinear spec syntax

alec-flexcompute:


tomflexcompute:
I think this needs to be merged into the develop branch instead.

